PR #18, opened 2020-04-01
feat: add option to use cucumber tags in run command

Mohammer5:
With this we can do the following:

`npx d2-utils-cypress run --tags "@highlighting and not @skip"`

It's worth mentioning that both features and scenarios can be highlighted.
With the following feature file, only the first scenario will be tested.
All other feature files and the scenarios in this feature file will be ignored:

```feature
@highlighting
Feature: Options can be highlighted

    Scenario: An option is highlighted
        # steps....

    @skip
    Scenario: An option is not highlighted
        # steps....
```

<hr />

In order to try/test this in an existing repo, I had to do this (because `concurrently` doesn't accept parameters that we can pass to one of the commands):

`"cy:run": "concurrently --kill-others --success first -n storybook,cypress 'yarn start:testing' 'wait-on http-get://localhost:5001 && d2-utils-cypress run --tags \"$TAGS\"'",`

and then call it like this:

`TAGS="@highlighting and not @skip" yarn cy:run`

### DIFF
--- a/src/commands/run.js
+++ b/src/commands/run.js
@@ -9,22 +9,29 @@ exports.aliases = ['r']
 exports.desc = 'Run Cypress tests'
 
 exports.builder = yargs =>
-    yargs.option('headed', {
-        describe: 'Run in headed mode',
-        type: 'boolean',
-        default: false,
-    })
+    yargs
+        .option('headed', {
+            describe: 'Run in headed mode',
+            type: 'boolean',
+            default: false,
+        })
+        .option('tags', {
+            describe: 'Use cucumber tags',
+            type: 'string',
+            default: '',
+        })
 
 exports.handler = argv => {
     log.info('d2-utils-cypress > run')
 
-    const { headed, port, browser } = argv
+    const { tags, headed, port, browser } = argv
 
     const opts = {
         mode: 'run',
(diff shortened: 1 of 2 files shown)


Mohammer5:
Ok.. for some reason when installing the local version of the lib the tests are not working. But that's regardless of what branch has been checked out. So it's not an issue these changes introduce. I can see that the correct feature files are used when using tags and this branch, but can't verify that the test actually run successfully.

@varl do you have an idea why I can't use the local version of the app? I ran `yarn build` in the `cli-utils-cypress` directory and then `yarn add -D file:../cli-utils-cypress` in `ui-core`.
When adding a console.log to the `src/tools/cypress.js` - printing the command and it's parameter - it looks 100% correct. It's as if the `cypress-cucumber-preprocessor` config file is ignored.

varl:
`yarn build` just builds the documentation in this repo. :)

varl:
I'll take a look, may have something to do how it evaluates the CWD.

Mohammer5:
Still relevant but blocked by refactor to monorepo

jenniferarnesen:
Superceded by https://github.com/dhis2/cli-utils-cypress/pull/90

PR #30, opened 2020-04-22
feat: add custom data-test selector syntax

Mohammer5:
This PR will add three things:

## 1. Extend functionality of the `get` and `find` commands

Having to write out the data test selector, especially when using several of them in one query, made the test code look quite bloated and much harder to parse when reading through it.
I've extracted this from the @dhis2/ui-core#639 branch, where I created a substitute syntax for writing selectors with data-test props, so these are all equivalent pairs:

```js
cy.get('[data-test="dhis2-uicore-transfer-sourceoptions"] [data-test="dhis2-uicore-transferoption"]')
cy.get('{transfer-sourceoptions} {transferoption}')

cy.get('[data-test="dhis2-uicore-transfer-filter"] input')
cy.get('{transfer-filter} input')

cy.get('.some-element').find('[data-test="dhis2-uicore-transfer-sourceoptions"] [data-test="dhis2-uicore-transferoption"]')
cy.get('.some-element').find('{transfer-sourceoptions} {transferoption}')

cy.get('.some-element').find('[data-test="dhis2-uicore-transfer"] input')
cy.get('.some-element').find('{transfer} input')

cy.get('[data-test="dhis2-appname-name"] [data-test="dhis2-appname-name2"]')
cy.get('{name} {name2}', { prefix: 'dhis2-appname' })

cy.get('.container [data-test="foobar-name"]')
cy.get('.container {foobar-name}', { prefix: '' })
```

**NOTE:** The original functionality of the two commands is kept in tact

## 2. Prompt the consumer to provide a prefix for the data test prop when installing

The extended `get` and `find` commands use the `dhis2_datatest_prefix` environment variable's value as default value for the prefix. In order to set this value, the install script prompts the user to supply a value. It'll suggest `dhis2-[lib/app-name]` where `lib/app-name` comes from the name in the package.json.

## 3. Expose the helper functions

This can be helpful when trying to get elements that might not exist without cypress to throwing a timeout, like this:

```js
// will yield a jQuery instance with length 0
cy.get('{transfer-pickedoptions}').then($pickedOptions => {
  return $pickedOptions.find(
    dataTestNameToSelector('transferoption')
  )
})
```

### DIFF
--- a/helper.js
+++ b/helper.js
@@ -1,0 +1,7 @@
+const dataTestNameToSelector = require('./helper/dataTestNameToSelector')
+const parseSelectorWithDataTest = require('./helper/parseSelectorWithDataTest')
+
+module.exports = {
+    dataTestNameToSelector,
+    parseSelectorWithDataTest,
+}
--- a/helper/dataTestNameToSelector.js
+++ b/helper/dataTestNameToSelector.js
@@ -1,0 +1,17 @@
+/**
+ * @param {string} dataTestName
+ * @param {string} [prefix] - Default to "dhis2-uicore"
+ * @returns {string}
+ */
+const dataTestNameToSelector = (dataTestName, prefix) => {
+    const defaultPrefix = Cypress.env('dhis2_datatest_prefix') || ''
+    // Empty string is a valid value, so check for undefined
+    const actualPrefix = typeof prefix === 'undefined' ? defaultPrefix : prefix
+    const dataTestId = actualPrefix
+        ? `${actualPrefix}-${dataTestName}`
+        : dataTestName
+
+    return `[data-test="${dataTestId}"]`
+}
+
+module.exports = dataTestNameToSelector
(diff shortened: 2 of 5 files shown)


amcgee:
Also I think as-is we might end up with `dhis2-@baocontacttracingapp` in the last case?

amcgee:
On the other hand, maybe we want to always prefix with `dhis2-` to differentiate from other types of `data-test` attributes, was that the intent?

Mohammer5:
@amcgee that's a great suggestion! `dhis2-@baocontacttracingapp` shouldn't happen.
Additionally, I'll replace `/^@[^\/]+\//` with an empty string to remove the namespace from the name and use that one (without the `@`) as a fallback value if there's no org name in the `package.json`

varl:
Pretty helpful addition! :+1: 

Are your additions done so this can be reviewed with @amcgee's comments in mind?

Mohammer5:
Yes, I was trying to add a couple more things, but I think it's better to do that in another PR

Mohammer5:
There are things I've done differently:
* The `-app` suffix is now removed
* The default org name is now `dhis2`, so `capture-app` for example will become `dhis2-capture` (I guess this tool will be used mostly in dhis2 contexts)

Mohammer5:
@amcgee ping

PR #34, opened 2020-05-01
feat: add concurrently & wait-on command

Mohammer5:
BREAKING CHANGE: The `d2-utils-cypress open` and `d2-utils-cypress run` now use `concurrently` and `wait-on` by default. If new behavior is undesired, it can be disabled with `--appStart ''`.

This PR does 3 things

1. refactors the `install` command's code
2. only writes the `testFiles` config to the `cypress.json` when using cucumber
3. Uses `concurrently` and `wait-on` when using the `open` and `run` commands

### DIFF
--- a/src/commands/install/createCypressConfig.js
+++ b/src/commands/install/createCypressConfig.js
@@ -30,7 +30,9 @@ const createCypressConfig = async force => {
         {
             baseUrl: cypressAnswers.baseUrl,
             video: cypressAnswers.video,
-            ...(cypressAnswers.projectId ? [cypressAnswers.projectId] : []),
+            ...(cypressAnswers.projectId
+                ? { projectId: cypressAnswers.projectId }
+                : {}),
         },
         force
     )
--- a/src/commands/install/createCypressConfig.js
+++ b/src/commands/install/createCypressConfig.js
@@ -1,6 +1,6 @@
 const inquirer = require('inquirer')
-
 const { write } = require('../../utils/fs.js')
+const { CYPRESS_CONFIG_PATH } = require('../../utils/paths.js')
 
 const createCypressConfig = async force => {
     const prompt = inquirer.createPromptModule()
@@ -11,12 +11,6 @@ const createCypressConfig = async force => {
             name: 'baseUrl',
             message: 'URL that Cypress should run tests against:',
             default: 'http://localhost:3000',
-        },
-        {
-            type: 'input',
-            name: 'testFiles',
-            message: 'Glob pattern for the test files to run:',
-            default: '**/*.feature',
         },
         {
             type: 'confirm',
@@ -32,10 +26,9 @@ const createCypressConfig = async force => {
     ])
 
     write(
-        'cypress.json',
+        CYPRESS_CONFIG_PATH,
         {
             baseUrl: cypressAnswers.baseUrl,
-            testFiles: cypressAnswers.testFiles,
             video: cypressAnswers.video,
             ...(cypressAnswers.projectId ? [cypressAnswers.projectId] : []),
         },
--- a/src/commands/install/createCypressConfig.js
+++ b/src/commands/install/createCypressConfig.js
@@ -1,0 +1,46 @@
+const inquirer = require('inquirer')
+
+const { write } = require('../../utils/fs.js')
+
+const createCypressConfig = async force => {
+    const prompt = inquirer.createPromptModule()
+
+    const cypressAnswers = await prompt([
+        {
+            type: 'input',
+            name: 'baseUrl',
+            message: 'URL that Cypress should run tests against:',
+            default: 'http://localhost:3000',
+        },
+        {
+            type: 'input',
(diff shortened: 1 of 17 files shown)


src/commands/open.js
```diff
@@ -1,25 +1,42 @@
...
+        const waitAndCypress = [waitOnCommand, cypressCommand].join(' && ')
+
+        concurrently(
+            [
+                { command: appStart, name: 'app' },
+                { command: waitAndCypress, name: 'cypress' },
+            ],
+            { kill: ['success'] }
+        )
+    }
```
amcgee, line 41: Very minor: could we factor this part out from this and the `run` command, since the code is identical?

src/commands/open.js
```diff
@@ -1,25 +1,42 @@
...
     }
 
-    cypress(opts)
+    if (!appStart) {
+        log.info('"appStart" empty. Running cypress directly')
+
+        cypress(opts)
+    } else {
+        const cypressCommand = cypress({ ...opts, exec: false })
+        const waitOnCommand = `npx --no-install wait-on ${waitOn}`
```
varl, line 31: This works great in apps, but when used through the global install of d2 (i.e. `d2 utils cypress`) this doesn't work for the same reasons it doesn't in `d2 style`.

I've been working on an alternative style that shouldn't get confused when e.g. `d2` is installed globally using Yarn vs. NPM (they have different `node_modules` globally, and `npx` looks in the NPM one and not Yarn one).

There's also the case of not hoisting dependencies (not talking about de-duping) that are _only_ used in e.g. `d2-style` which this "inside out" approach allows to work. Needs more testing and bug fixing but it's in `@dhis2/cli-style@alpha` right now: https://github.com/dhis2/cli-style/commit/fd620c56ca8798ab0e84366d97cdbd0d0fdaba03

If it works out, maybe we should move the `cli-style/src/utils/run.js` file to e.g. `cli-helpers-engine` and use that as the primary process runner.

varl: What I'm saying is: use `npx` for now, and be aware that there are some gotchas that we want to resolve, and there are experiments for that in `dhis2/cli-style` going on.

src/commands/install/createCypressConfig.js
```diff
@@ -0,0 +1,39 @@
...
+            message: 'The unique Cypress project ID:',
+        },
+    ])
+
+    write(
+        CYPRESS_CONFIG_PATH,
+        {
+            baseUrl: cypressAnswers.baseUrl,
+            video: cypressAnswers.video,
+            ...(cypressAnswers.projectId ? [cypressAnswers.projectId] : []),
```
Mohammer5, line 33: @varl This should be `cypressAnswers.projectId ? { projectId: cypressAnswers.projectId } : {}`, shouldn't it?
I'll test on Tuesday what'll happen when I provide a value for projectId, guess none of use actually tried to use this one so far.

Which leads me to another thought (cc @amcgee): Would it make sense to have integration tests for our cli libraries (i. e. creating a temp dir, `yarn init`, `d2 utils cypress install`, answer the prompts depending on what the test wants to test and finally check the result)?

I'm not sure tests are actually worth the time here. I guess once our CLI tools work, we kind of know that they work, they're not used in production apps anyways so if a bug exists, it's most like going to be nothing but a nuisance.

src/commands/open.js
Mohammer5: Good to know. So if I get it right `d2 utils cypress open` won't work when using the globally installed cli utility. Never thought about that use-case before :sweat_smile:

src/commands/open.js
Mohammer5: Should look much better now

src/commands/open.js
amcgee: I think `npx` is probably OK here for the moment because `wait-on` doesn't need to find other dependencies (like `eslint` does when resolving plugins from the .eslintrc file).  So should be fine for now but worth moving to a better process runner when we have it.

src/commands/install/createCypressConfig.js
amcgee: @Mohammer5 I think that would be nice, at least for smoke tests, but I don't think it's highest priority as you mentioned - if things break, we will notice and can quickly fix them without much fuss since cli tools are continuously deployed

amcgee: But when we have some bandwidth I do think this is a very good idea, especially for tools that have implications for external developers making DHIS2 apps because bugs introduce friction there

amcgee: @Mohammer5 yeah I think this line `...(cypressAnswers.projectId ? [cypressAnswers.projectId] : [])` is incorrect though, needs to be an object otherwise it will set `0: cypressAnswers.projectId`

Mohammer5: Fixed

Mohammer5:
This should be ok now

PR #37, opened 2020-05-12
refactor: transform to monorepo with cli & commands package

Mohammer5:
The reasons for changing this into a monorepo are:

1. separate cli utility & test-run support functions
2. being able to write the support functions with es6+ `import` instead of keeping them in one file

<hr />

* This means that the version released once this is merged won't contain the support functions by default, which makes this a breaking change.
  -> I've added an extra step in the `install` script which will install the `@dhis2/cypress-commands` dev dependency (can be found in [`packages/cli/src/commands/install/createCypressSupportFile.js`](https://github.com/dhis2/cli-utils-cypress/blob/Monorepo_and_commands_package/packages/cli/src/commands/install/createCypressSupportFile.js))
* I'd like to add the cli library as a dev dependecy as well during the `install` process as it's required to run the cypress commands. But I'd prefer to do that in another PR

BREAKING CHANGE: Adopting to this change requires manual installation of the @dhis2/cypress-commands package


Mohammer5:
@amcgee ping

amcgee:
> I've tried to move the current cli package to packages/utils and created a second one in packages/commands

In other situations where we have a split like this (i.e. `app-platform`) we call the CLI modules `cli`, seems like a more descriptive name than `utils`?  Also, should `commands` maybe be `support`?  I don't feel as strongly about that one.

> I'm not sure how to manage the dependencies, my guess is that only dependencies shared between the sub packages should be moved to the main package.json?

Typically dependencies should live with the package that needs them.  Dev dependencies can optionally be omitted and moved up to the top-level package.json (see [`app-platform` for one example](https://github.com/dhis2/app-platform/blob/master/package.json), not necessarily the best one)

Mohammer5:
I'd prefer "commands", all the officially supported plugins use that name as well: https://docs.cypress.io/plugins/index.html

Will rename `utils` to `cli`

package.json
```diff
@@ -1,48 +1,36 @@
...
+    "cli:lint": "cd ./packages/cli && yarn lint",
+    "cli:lint:js": "cd ./packages/cli && yarn lint:js",
+    "cli:lint:text": "cd ./packages/cli && yarn lint:text",
+    "cli:lint:staged": "cd ./packages/cli && yarn lint:staged",
+    "cli:format": "cd ./packages/cli && yarn format",
+    "cli:format:js": "cd ./packages/cli && yarn format:js",
+    "cli:format:text": "cd ./packages/cli && yarn format:text",
+    "cli:format:staged": "cd ./packages/cli && yarn format:staged",
+    "cli:build": "cd ./packages/cli && yarn build",
+    "cli:start": "cd ./packages/cli && yarn start"
```
amcgee, line 34: While these are technically fine, they're quite repetitive and need to be synced with the scripts in each workspace package.  Yarn allows you to run a command in an individual workspace like this `yarn workspace @dhis2/cli-utils-cypress format:stages` (unfortunately it matches on the package name, not the directory) or on all workspaces with `yarn workspaces format:staged`.  Maybe it's enough to just have these scripts at the root?

```js
"build": "yarn workspaces build",
"start": "yarn workspace @dhis2/cypress-helpers build && yarn workspace @dhis2/cli-utils-cypress start",
"format": "yarn workspaces format",
"lint": "yarn workspaces lint"
```

Or something similar?

packages/commands/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@dhis2/cypress-helpers",
```
amcgee, line 2: I think this should probably match the workspace directory, either `@dhis2/cypress-helpers` in `packages/helpers` or `@dhis2/cypress-commands` in `packages/commands`

packages/commands/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "version": "1.0.0",
+  "description": "Cypress commands for dhis2 related apps",
+  "main": "build/cjs/lib.js",
+  "module": "build/es/lib.js",
+  "repository": "https://github.com/dhis2/cypress-helpers",
```
amcgee, line 7: This should point to this repository, `dhis2/cypress-helpers` doesn't exist.

You can use the last form of the `repository` field [documented here](https://docs.npmjs.com/files/package.json#repository) which allows you to specify a folder within the repo

packages/cli/package.json
```diff
@@ -0,0 +1,47 @@
+{
+    "name": "@dhis2/cli-utils-cypress",
+    "description": "DHIS2 Cypress configuration and utilities",
+    "repository": {
```
amcgee, line 4: You can use the last form of the `repository` field [documented here](https://docs.npmjs.com/files/package.json#repository) which allows you to specify a folder within the repo

amcgee:
Also important - all the dotfiles and github configuration should stay at the root level, it doesn't work when it's in the `packages/*` directories (many are currently duplicated)

Notably:
- `.github/*` - this is why the `check` action isn't running.  The action also needs to call a `yarn` script that exists in the root `package.json`
- `.dependabot`
- `CHANGELOG.md`, `README.md`
- `.editorconfig`, `.eslintrc.js`, `.huskyrc.js`, `.prettierrc.js`
- `netlify.toml` (needs to be updated unless `docs` is moved to the root)
- `.browserlistrc` - can probably just be removed

Also - a `yarn.lock` file should only be necessary at the root, the other ones can be removed

And finally - it's nice to see that `d2-app-scripts` works here, though it's not designed for Cypress support compilation and does probably several things (like translation extraction) that are superfluous or could lead to weird bugs.  Probably fine for now, but we should think about having a separate build setup for Cypress support maybe?

amcgee:
The `start` script in `commands` package is probably unnecessary and uses `nodemon` which can be removed.  You can just run `yarn build --watch` (which runs `d2-app-scripts build --watch` under the hood)

amcgee:
The lint and format scripts can probably all just move to the root rather than duplicating them in both packages...?

Mohammer5:
> templates/support.js and templates/plugins.js were moved but were not updated - they still have const xyz = require('@dhis2/cli-utils-cypress/xyz') when they should be importing from the new helpers package.

The `plugins` one currently still lives in the cli package, so that one should be fine. I'm wondering about the other one though. It would only make sense to update that one if we installed the commands package in the application when running `d2 utils cypress install`.

Currently the install script doesn't install the `@dhis2/cli-utils-cypress` dep either, which is needed for running cypress.

What I currently have in mind is: The `install` script should add the `@dhis2/cli-utils-cypress` dep and add some scripts to the package.json and either we don't add the commands at all or add the dependency during the install scripts (although we'd have to know whether the consumer is using `yarn` or `npm`, I don't know if we support 3rd party platform apps with npm).

> Minor: cli-style should be moved to the root package.json since it's needed in both packages as a dev dependency.

Done

> Minor: cli-utils-docsite should probably be generated from the root rather than packages/cli/docs? Also the API page isn't generated (it should be using jsDoc) but that was probably an existing issue.

Done

> Minor: it surprised me to see start serve the docs, I would usually expect that to start a sample application instead (see https://github.com/dhis2/app-platform/blob/master/package.json#L27-L30 for one approach)

There's no example app needed in this case, but I've renamed the docs commands

> And finally - it's nice to see that d2-app-scripts works here, though it's not designed for Cypress support compilation and does probably several things (like translation extraction) that are superfluous or could lead to weird bugs. Probably fine for now, but we should think about having a separate build setup for Cypress support maybe?

I think for now it's fine, we can always change it when needed

packages/commands/package.json
Mohammer5: I've renamed both the package name as well as the folder to `cypress-commands`

package.json
Mohammer5: I've moved them all up to the root

amcgee (changes requested):
Looking good!  Made a few minor suggestions

packages/cli/src/commands/install/createCypressSupportFile.js
```diff
@@ -0,0 +1,121 @@
...
+            {
+                type: 'input',
+                name: 'packageManager',
+                message: `This step will install the @dhis2/cypress-commands package.\nWhich package manager are you using? Yarn [${YARN}], Npm [${NPM}], skip [${NONE}], (Default: ${YARN})`,
+                default: YARN,
+            },
+        ])
+
+        packageManager = userInput.packageManager
+    } while (![YARN, NPM, NONE].includes(packageManager))
```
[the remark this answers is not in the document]
Mohammer5: I chose to use the `list` type, was quicker, can still be changed later on

Mohammer5:
@amcgee addressed it all the things you found

amcgee (approved):
Changes LGTM, thanks @Mohammer5!

amcgee:
Any reason the `check` action isn't running?

varl:
> Any reason the `check` action isn't running?

There is no `.github/workflows` in this branch.

varl:
Looks good overall!

.github/workflows/node-lint.yml
```diff
@@ -1,21 +0,0 @@
-name: 'dhis2: lint (node)'
```
varl, line 1: These need to be re-added.

.github/workflows/node-publish.yml
```diff
@@ -1,45 +0,0 @@
-name: 'dhis2: publish (node)'
```
varl, line 1: And this.

packages/cli/package.json
```diff
@@ -0,0 +1,37 @@
...
+        "cypress": "^4.0.2",
+        "cypress-cucumber-preprocessor": "^2.0.1",
+        "fs-extra": "^8.1.0",
+        "inquirer": "^7.0.4",
+        "wait-on": "^4.0.2"
+    },
+    "publishConfig": {
+        "access": "public"
+    },
+    "devDependencies": {
```
varl, line 33: `devDeps` are not necessary for packages, it's preferable to only have them in the root `package.json`.

packages/cypress-commands/src/commands/login.js
```diff
@@ -0,0 +1,57 @@
+const LOGIN_END_POINT = 'dhis-web-commons-security/login.action'
```
varl, line 1: There was a discussion about this on Slack that I didn't catch the conclusion of, but `dhis-web-commons-security/login.action` gives me a redirect to `dhis-web-commons/security/login.action`. Could we double check that this is indeed the correct URL to use here?

amcgee: This is the correct endpoint-

`dhis-web-commons-security` is a Spring API endpoint which only accepts POST requests

`dhis-web-commons/security` is the STRUTS login app (just the UI, it POSTs to Spring on login)

amcgee: Definitely room for improvement there, which we will address when replacing the web portal

varl: Perfect, thanks. Glad to have it documented here in the PR. :+1:

jenniferarnesen: the `web-commons-security` is a POST only endpoint to be used for authentication. The `web-commons/security` is for the old Struts framework. So my understanding is that the former is the correct one for our apps to authenticate.

jenniferarnesen: https://dhis2.slack.com/archives/C1XNHEY0L/p1603723486038300

Mohammer5:
@varl I've readded the github workflows. Wondering about the publish workflow as there are two individual packages now, do we need to adjust anything in that one or on the individual packages?

varl:
> @varl I've readded the github workflows. Wondering about the publish workflow as there are two individual packages now, do we need to adjust anything in that one or on the individual packages?

Thanks. The publish workflow is monorepo aware, no changes needed.

varl (approved):

PR #78, opened 2020-10-16
fix: correct url and needed props for login command

jenniferarnesen:
Fixes include:
* correct login url
* add `form` and `followRedirect` properties
* remove the auth header since it isn't needed for logging in


Mohammer5 (approved):

vilkg (commented):
Lgtm

PR #82, opened 2020-10-26
feat: expose the spec config option

jenniferarnesen:
The --spec option allows for specific test files to be run. This can be useful for running a set of smoke tests on ci, or running another subset of the tests.

### DIFF
--- a/src/commands/open.js
+++ b/src/commands/open.js
@@ -9,8 +9,8 @@ exports.builder = yargs =>
     yargs.option('appStart', appStart).option('waitOn', waitOn)
 
 exports.handler = argv => {
-    const { appStart, port, browser, waitOn } = argv
-    const cypressOptions = { mode: 'open', browser, port }
+    const { appStart, port, browser, waitOn, spec } = argv
+    const cypressOptions = { mode: 'open', browser, port, spec }
 
     log.info('d2-utils-cypress > open')
     execCypress({ appStart, waitOn, cypressOptions })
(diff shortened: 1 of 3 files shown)


Mohammer5:
I'm not sure this is the best solution to the problem because it doesn't work well with `concurrently`. I came up with a solution for this a while ago: dhis2/cli-utils-cypress#18

It hasn't been merged yet though. I would still consider adding this pattern to the library as it can make sense to use it this way manually. Did the changes to the library work? I'm wondering about potential issues with the paths (with cypress being a dependency of a dependency), it could work though

jenniferarnesen:
> It hasn't been merged yet though. I would still consider adding this pattern to the library as it can make sense to use it this way manually. Did the changes to the library work? I'm wondering about potential issues with the paths (with cypress being a dependency of a dependency), it could work though

It did work, yes. (modified my node_modules/@dhis2/cli-utils-cypress src code)

jenniferarnesen:
> I'm not sure this is the best solution to the problem because it doesn't work well with `concurrently`. I came up with a solution for this a while ago: #18

I like the tag solution better - more flexible.

HendrikThePendric:
@jenniferarnesen @Mohammer5 it looks like we can close this?

Mohammer5:
> I would still consider adding this pattern to the library as it can make sense to use it this way manually.

I think this is still valid. Especially for cypress implementations that don't use cucumber

HendrikThePendric:
I think this one can be closed now, since we have removed the `open` and `run` command of the CLI tool and are now using cypress directly.

PR #91, opened 2020-11-03
chore: update semantic commit rules to allow merge commits

jenniferarnesen:
This is in line with other DHIS2 repositories

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,4 +1,4 @@
-titleOnly: true
+titleOnly: false
 commitsOnly: false
-titleAndCommits: false
-allowMergeCommits: false
+titleAndCommits: true
+allowMergeCommits: true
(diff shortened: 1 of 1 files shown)


varl:
Great, the master configuration installed with new versions of `cli-style` is:

```
titleOnly: false
commitsOnly: false
titleAndCommits: true
allowMergeCommits: true
```

Any reason to not use that?

jenniferarnesen:
> Great, the master configuration installed with new versions of `cli-style` is:
> 
> ```
> titleOnly: false
> commitsOnly: false
> titleAndCommits: true
> allowMergeCommits: true
> ```
> 
> Any reason to not use that?

Sure if that fixes the broken build on https://github.com/dhis2/cli-utils-cypress/pull/90, then I'm happy :)

varl (approved):

PR #92, opened 2020-11-04
fix(run command): make command exit correctly

Mohammer5:
Fixes TECH-472

This will make sure that the `d2 utils cypress run` command exits properly.
The previous `concurrently` configuration didn't cause the process to exit when the `run` process has finished in some cases

### DIFF
--- a/packages/cli/src/tools/waitOnAndCypress.js
+++ b/packages/cli/src/tools/waitOnAndCypress.js
@@ -1,5 +1,6 @@
-const { getCypressCommand } = require('./getCypressCommand.js')
+const log = require('@dhis2/cli-helpers-engine').reporter
 const concurrently = require('concurrently')
+const { getCypressCommand } = require('./getCypressCommand.js')
 
 exports.waitOnAndCypress = ({ appStart, waitOn, cypressOptions }) => {
     const { cmd, options } = getCypressCommand(cypressOptions)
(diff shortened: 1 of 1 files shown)


varl (commented):
Interesting, everything seems to work as before. :+1: 

Do you know of a way to reliably reproduce the original problem? I haven't been able to verify this fix since I don't understand under what circumstances the exit fails.

Mohammer5:
I haven't found a reliable way to reproduce this yet. I'll try to find out the exact cause

varl:
So I haven't seen a problem with this PR yet, so wondering if you want to merge straight to `master` or take it for a spin on `next` before releasing it?

Mohammer5:
I'm still wondering why this was necessary. I recall that @HendrikThePendric had an issue with this once and I had it a couple of times. I think it's safe to merge this one, but i'd also like to figure out why this is necessary in the first place. Should I create an issue (there's a chance it's so low in priority it'll never be investigated tho)?

varl:
Let's wait with merging this at all then, and if it pops up again, we investigate.

Mohammer5:
:ok_hand: Closed until relevant

varl:
I can reproduce this locally and consistently.

varl (approved):
This fixes it !

PR #101, opened 2020-11-19
feat: autoregister commands & add enableAutoLogin setup

Mohammer5:
Fixes CLI-14

* [x] Autoregister all commands
* [x] Add `enableAutoLogin` setup helper
* [x] ~~Commit [33d7a6d](https://github.com/dhis2/cli-utils-cypress/pull/101/commits~~/33d7a6dab7562be5df57003a302606b9063936b3) (`feat(login): add command to fill login form`) is a bit questionable
  * **REMOVED**
  * Right now I'd prefer to not add a command for this at all
  * But I can see that some apps will have to fill in form data
  * We could either drop this command (by removing the commit) or keep it


packages/cypress-commands/src/commands/fillAndSubmitLoginForm.js
```diff
@@ -0,0 +1,19 @@
...
+ * @returns {void}
+ */
+export const fillAndSubmitLoginForm = ({ loginUrl, userName, password }) => {
+    loginUrl && cy.get('#server').type(loginUrl)
+    cy.get('#j_username').type(userName)
+    cy.get('#j_password').type(password)
+    cy.get('{loginsubmit}', { prefix: 'dhis2-adapter' }).click()
+}
+
+Cypress.Commands.add('fillAndSubmitLoginForm', fillAndSubmitLoginForm)
```
paschalidi, line 19: Hey, I know this is being added for the capture app use-case. Backstory time: sometime ago we had this bug that Gintare, Jeniffer and others solved in the `login` helper of the utils. Basically we couldnt login by posting a request to the server as a workaround we used the ui to login.. 

That said and if there isnt any other use cases for this utility perhaps you can think of removing it.

packages/cypress-commands/src/index.js
```diff
@@ -1,6 +1,21 @@
...
 
-// commands
-export { registerCommands } from './commands/registerCommands'
+// setup helpers
+export { enableAutoLogin } from './setups/enableAutoLogin'
+
+// backward compatibility
+export const registerCommands = () => {
+    cy.log('The usage of `registerCommands` has been depricated')
+    cy.log('This function does nothing anymore and can be removed')
```
Mohammer5, line 19: `... doesn't do anything anymore ...` ?
`This function does nothing and can be removed` ?

HendrikThePendric: I agree.

packages/cypress-commands/src/commands/fillAndSubmitLoginForm.js
HendrikThePendric: I agree with @paschalidi: if `loginBasicAuth` works reliably, then I'd classify this command as redundant.

Also, if there are compelling reasons for keeping it, I think we should rename it. My qualm with the current name is that it doesn't include a reference to "platform app". I think it should do though, since this login form is specific to the app-shell.

packages/cypress-commands/src/commands/login.js
```diff
@@ -53,5 +53,14 @@ const loginDev = () => {
  * https://docs.cypress.io/guides/guides/web-security.html#One-Superdomain-per-Test
```
HendrikThePendric, line 53: I think this comment has become misplaced, it probably should be above `loginBasicAuth`?

packages/cypress-commands/src/commands/login.js
```diff
@@ -53,5 +53,14 @@ const loginDev = () => {
...
  */
 export const login = () => {
+    ;[
+        'DEPRICATION WARNING:',
+        'The `login` function will is an older version.',
+        'It is left in place to prevent breakage.',
+        'Please use either the `loginThroughForm` command or `autoLogin` setup.',
+    ].forEach(message => cy.log(message))
+
     return loginDev()
```
HendrikThePendric, line 63: This ties into my other comments. I think we should get rid of the warning and just keep the command. Ideally, we should just use the `loginBasicAuth` approach at all time (i.e. just implement the `login` command as `loginBasicAuth`). If one approach to logging in works reliably we should use that and keep things simple.

What are the reasons for keeping both approaches around?

HendrikThePendric (changes requested):
This will be very handy, great work! I've left a few comments in the code, most relate to the same thing so I'll also try to explain my general ideas below:
- In my opinion, if possible, we should only have one `login` command and ideally this should be doing what is now `loginBasicAuth`, since that is independent of the app-platform login modal.
- And besides that `login` command we should have the `enableAutoLogin` setup, which calls `cy.login()` in the `before` hook and in the `beforeEach` it preserves the cookie and localStorage item. Probably I would prefer a slightly different name, i.e. `enablePersistedSession` or sth.

packages/cypress-commands/src/setups/enableAutoLogin.js
```diff
@@ -0,0 +1,40 @@
...
+        throw new Error(
+            'No `dhis2_base_url` found. Please make sure to add it to `cypress.env.json`'
+        )
+    }
+
+    before(() => {
+        // Persist this across tests so we don't have to login before each test
+        Cypress.Cookies.defaults({
+            whitelist: 'JSESSIONID',
+        })
```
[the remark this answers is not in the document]
HendrikThePendric: Oh, and in future version, `whitelist` is going to get changed to `preserve` see https://github.com/cypress-io/cypress/pull/7782

packages/cypress-commands/src/commands/login.js
Mohammer5: backward compatibility

packages/cypress-commands/src/index.js
Mohammer5: with which of the three alternatives?

packages/cypress-commands/src/commands/login.js
varl: A deprecation warning with an intent to remove in a future version is considerate towards users of the lib.

packages/cypress-commands/src/index.js
HendrikThePendric: Sorry, I misinterpreted things... Get it now though... I think it's important to mention all three aspects actually, but I think we can provide that info in one sentence perhaps:
```
The usage of `registerCommands` has been deprecated and is now a no-op, because commands are registered automatically.
```

packages/cypress-commands/src/commands/login.js
HendrikThePendric: Sorry but I think we have a misunderstanding here. Maybe I am missing something, but this is the situation as I've understood it:
- The original `cy.login` did not work properly and people have started building there own.
- This has produced another way of logging in, through the UI. This new way of logging in would now be introduced into this lib as `fillAndSubmitLoginForm`. However, as both @paschalidi and I have tried suggesting: we probably don't need this new method of logging in.
- So my conclusion was that we just need one command that we can use to reliably login, and which is not dependent on a UI that is only present in platform-apps.

So assuming the reasoning above is correct we shouldn't deprecated `cy.login`, but just fix it instead.

That was basically the main point I was trying to make in several of my comments and my review summary as well, see https://github.com/dhis2/cli-utils-cypress/pull/101#pullrequestreview-537370270. But I guess it wasn't very clear.

If things still aren't clear or I'm missing the point why my suggestion is not viable, let's discuss in the meeting we have together tomorrow @Mohammer5

varl: Alright, I misunderstood something between here and there. I thought that both commands work and then it makes sense to deprecate the less preferred on.

Mohammer5: @HendrikThePendric and I talked about this. We agreed that we don't need to deprecate this command, it does virtually the same as the `enableAutoLogin` without the hooks, so we can just update this command, and use it inside the setup function

packages/cypress-commands/src/commands/fillAndSubmitLoginForm.js
Mohammer5: will be removed as it's an unnecessary abstraction. Apps can do it themselves

packages/cypress-commands/src/setups/enableAutoLogin.js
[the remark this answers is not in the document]
Mohammer5: will be handles in a different PR when upgrading the cypress version

Mohammer5:
@HendrikThePendric I've updated the login command, the setup file and the deprecation warning

HendrikThePendric (approved):
Thanks for addressing the points we discussed. Great work!

PR #102, opened 2020-11-20
feat(codemods): add codemod for transforming get/find

Mohammer5:
Fixes CLI-12

* [x] Adds a codemod that will transform `get`/`find` to `getWithDataTest`/`findWithDataTest`
* [x] Adds a test for the codemod

How to test:
* Clone the [codemods](https://github.com/dhis2/codemods) repository (`git clone https://github.com/dhis2/codemods`)
* Checkout the branch `git checkout CLI-10_Turn_into_cli_utility_V2`
* Install the deps (`yarn`)
* Copy the test dir with the source files for the codemods test somewhere
```
cp \
  /path/to/cli-utils-cypress/packages/cypress-commands/codemods/__test__/replace-get-find-with-xxxxGetWithDataTest/get-find-source \
  /path/to/destination
```
* Use the binary of the codemods
```
./path/to/codemods/bin/d2-utils-codemods apply /path/to/destination \
  --codemodPath /path/to/cli-utils-cypress/packages/cypress-commands/codemods/replace-get-find-with-xxxxGetWithDataTest.js
```


varl:
I don't think the `yarn.lock` file should be in the cypress-commands folder, it should only be on the root project level.

Mohammer5:
@varl I've removed the file.

---

@dhis2/team-platform Let's talk about the name of the new helpers, currently they've:
* getWithDataTest
* findWithDataTest

I don't like the names too much because their intent is not really clear. They basically wrap `get` and `find` and handle the custom data-test syntax, but I find it very hard to come up with a proper name that't not 100000 characters long

ismay:
>    getWithDataTest
>    findWithDataTest

I think that that's fine 👍

varl:
`getByDataTest` and `findByDataTest` is perhaps slightly more descriptive, and it matches document methods like `getElementsByClassName` and `getElementById`, thought I think that those functions are better named because they state _what_ you are getting (elements).

Mohammer5:
The issue with `findByDataTest` is that you're not finding something by using a data test value, you're just finding with an extra selector syntax, then I'd prefer `With` over `By`

varl:
> The issue with `findByDataTest` is that you're not finding something by using a data test value, you're just finding with an extra selector syntax, then I'd prefer `With` over `By`

Ah, that's an important distinction. :thinking:

Mohammer5:
Maybe the initial definition is good enough.. At least there's documentation now (check the readme in the master branch)

varl (approved):
LGTM

PR #105, opened 2020-12-01
docs: add docs for everything

Mohammer5:
Just run `yarn docs:start` and check the docs in the browser


docs/getting-started.md
```diff
@@ -14,33 +21,24 @@ yarn d2-utils-cypress
...
-We are going to use `d2-utils-cypress` in this document, so prefix the
-commands with your preference.
+## Install the cypress commands
 
-# Setup Cypress for your project
-
-The CLI module provides an `install` command, that will set up some
-prerequisites for Cypress, and guide you through generating a
-configuration and a local environment configuration file.
+When using the cli tool's `install` command, this will be handled for you. If
```
HendrikThePendric, line 26: I think it'd be good to also provide a copy-paste command here.

docs/guides.md
```diff
@@ -0,0 +1,4 @@
+# Guides
+
+You can find a list of guides in the sidebar. They contain a step by step guide
+for different topics
```
HendrikThePendric, line 4: Add `.` (full stop)

docs/guides/enable-auto-login.md
```diff
@@ -0,0 +1,53 @@
...
+
+Here is an example what the environment file could look like
+
+```json
+{
+    "dhis2_base_url": "",
+    "dhis2_username": "",
+    "dhis2_password": ""
+}
+```
```
HendrikThePendric, line 53: I made a comment about this topic earlier, but the additional info I requested for the `cy.login` command is actually here already. Perhaps it would make sense to colocate the info about the env variables with the `cy.login` command and reduce the amount of info you are providing here. You could just explain that `enableAutoLogin` calls the `cy.login` command in the before hook and and makes sure the session is persisted in the `beforeEach` hook?

docs/getting-started.md
HendrikThePendric: Actually, I think this could be seen as duplicate info. What you have written in `docs/guides/setting-up-cli-tool.md` is more complete.

HendrikThePendric (approved):
Looking great! Left a few suggestions in the code.

docs/guides/enable-auto-login.md
HendrikThePendric: I was thinking about this one a bit more... You've mentioned two ways to set env variables, but there are actually 6 ways to do it. I think we should not include any instructions in this repo, but simply link to https://docs.cypress.io/guides/guides/environment-variables.html

Mohammer5: > Perhaps it would make sense to colocate the info about the env variables with the cy.login command and reduce the amount of info you are providing here.

It's important to note that this is a guide, not api docs, so I think the duplication is actually good. It should be part of a guide, not putting it into the docs to avoid duplication would be bad as the docs would be incomplete.

I'll add a link to all the possible ways to provide the env vars to the login api docs

> I think we should not include any instructions in this repo

I disagree. The reason why I've put this into the guide is to specifically inform the reader about git-ignoring the `cypress.env.json` so they can have their own backend instance and login credentials, while the data in `cypress.json` could be used for CI

docs/getting-started.md
Mohammer5: I've linked to the guide from this document, that should resolve this issue?

docs/guides/enable-auto-login.md
HendrikThePendric: > It's important to note that this is a guide, not api docs, so I think the duplication is actually good. It should be part of a guide, not putting it into the docs to avoid duplication would be bad as the docs would be incomplete.

💯 

> I disagree. The reason why I've put this into the guide is to specifically inform the reader about git-ignoring the `cypress.env.json`

Good point, very true. But maybe you could rephrase a bit then: first just explain how to set these vars (offer a link to the cypress docs), and then explain that security might be an issue. I don't think putting variables in the `cypress.json` is the only way to leak secrets. Other possibilities are:
- Some CI config file that defines env vars. To avoid this you would have to use some secrets there
- A dev might define some script in `package.json` that sets an env variable before it runs a command

So maybe some general words of caution would be good, plus perhaps some specific instructions about using a `cypress.env.json` file. Come to think of it, if you are mentioning gitignoring this file, perhaps you should also outline an approach to make sure the correct vars are available in CI too?

Mohammer5: Included links for setting up github action env vars and encrypted secrets

PR #106, opened 2020-12-02
feat: introduce network shim

HendrikThePendric:


packages/cypress-commands/src/setups/enableNetworkShim/captureRequests.js
```diff
@@ -0,0 +1,135 @@
...
+    requestStub.count += 1
+    state.duplicates += 1
+    const isNotModified = responseStatus === 304
+
+    if (!requestStub.nonDeterministic) {
+        // if requestStub.responseBody equals newResponseBody this is a simple
+        // duplicated request and we don't take any action
+        // And we deal with 304s the same way
+
+        if (requestStub.responseBody !== newResponseBody && !isNotModified) {
```
Mohammer5, line 92: I am confused about this condition.
In order to get to this line, a matching/existing response has to be found (in `findMatchingRequestStub.js`). Inside that function, there's a condition that compares the response bodies and only returns true if they're equal.
How can this case be happening then?

packages/cypress-commands/src/setups/enableNetworkShim/createStateFromFixtures.js
```diff
@@ -0,0 +1,45 @@
+import { isStubMode, getNetworkFixturesDir } from './utils'
+
+function parseFixtureFiles(fileNames) {
+    return cy
+        .all(
+            ...fileNames.map(fileName => () =>
+                cy.readFile(`${getNetworkFixturesDir()}/requests/${fileName}`)
```
Mohammer5, line 7: I would change this to `cy.fixture`. On one side I had to read the docs first to know that this function does more than its name reveals, it's also parsing JSON. On the other side, as these files are in the fixtures folder, I think `.fixture` is more appropriate

packages/cypress-commands/src/setups/enableNetworkShim/createStateFromFixtures.js
```diff
@@ -0,0 +1,45 @@
...
+        const config = {
+            serverMinorVersion,
+            hosts,
+            fixtureMode,
+            staticResources,
+            mode: Cypress.env('dhis2_api_stub_mode'),
+        }
+
+        return cy
+            .readFile(`${getNetworkFixturesDir()}/summary.json`)
```
Mohammer5, line 29: Same as above

packages/cypress-commands/src/setups/enableNetworkShim/index.js
```diff
@@ -0,0 +1,53 @@
...
+        if (isCaptureMode()) {
+            validateVersionMinor()
+        }
+        createStateFromFixtures({
+            hosts,
+            fixtureMode,
+            staticResources,
+        }).then(state => {
+            cy.wrap(state).as('networkShimState')
+        })
```
Mohammer5, line 33:
```js
        createStateFromFixtures({
            hosts,
            fixtureMode,
            staticResources,
        }).as('networkShimState')
```

packages/cypress-commands/src/setups/enableNetworkShim/stubRequests.js
```diff
@@ -0,0 +1,60 @@
...
+            try {
+                stubRequest(state, request)
+            } catch (error) {
+                console.error('NetworkShim stub error', error)
+            }
+        })
+    })
+}
+
+function stubRequest(state, request) {
```
Mohammer5, line 20: What happens if changes caused there to be a new, not yet captures, request? It seems like this case is not covered

packages/cypress-commands/src/setups/enableNetworkShim/utils.js
```diff
@@ -0,0 +1,96 @@
...
+        const isMatchingRequest =
+            path === r.path &&
+            method === r.method &&
+            (requestBody === r.requestBody ||
+                JSON.stringify(requestBody) === JSON.stringify(r.requestBody))
+
+        return isStaticResource
+            ? isMatchingRequest
+            : isMatchingRequest && testName === r.testName
+    })
```
Mohammer5, line 93: I'm generally against abbreviations, they just increase the cognitive overhead when reading the code

Mohammer5 (changes requested):
I'm generally having issues understanding what exactly `nonDeterministic` means in this context. Could you explain that (and maybe add that explanation to some kind of docs)?

---

What i miss is jsdocs for the various functions. It would help a great deal if I could also read the input parameter's args and the expected return type

packages/cypress-commands/src/setups/enableNetworkShim/captureRequests.js
```diff
@@ -0,0 +1,135 @@
...
+        })
+    })
+}
+
+async function captureRequest(state, request, response) {
+    const { host, path } = splitHostAndPath(request.url, state.config.hosts)
+    if (!host) {
+        // pass through
+        return response
+    }
```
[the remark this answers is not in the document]
HendrikThePendric: From a practical perspective:
Suppose an app does a request to some CDN, or more generally a public server that would be accessible during the CI e2e run, and you just want the request to happen naturally. That's what this condition is for.

and from a technical perspective:
- The NetworkShim only captures requests to domains specified in the `hosts` array.
- So if during a test run you are doing a request to `https://placeholderimages.com`, but that 
domain is not in the `hosts` array, requests to that url will not be captured or stubbed.

packages/cypress-commands/src/setups/enableNetworkShim/captureRequests.js
HendrikThePendric: A valid question, as this is one of the more complex parts of the beast. I'll try to explain:
- We start by looking if a matching request stub is already present. That means we find a match on `path`, `method` and `requestBody`
- But once we have found this we need to look further because actually three things could be happening here:
    - The response body matches the response body (or possibly one of the response bodies) on the stub.: this simply is a repeating request.
    - The response body does not match the stub: this means the request is _nonDeterministic_, an repeating request does not yield the same response body (I will add docs with definitions).
    - The response body is empty and the response status is 304: this is the server telling the browser that it should return its cached response. So this is a repeating request, but we threat this as an exceptional case: if we were to allow the normal logic to process the response body it would conclude that the response has changed (i.e. from some sort of array/object/string into an empty string). So it would essentially be treated as a nonDeterministic response, while it is the exact opposite. So for 304 responses we basically need to just look at last know response body and use that.

I hope this helps. It's not just a difficult part to understand but also to explain 😊 .

Here's some info about 304s:
> A 304 response cannot contain a message-body; it is always terminated by the first empty line after the header fields.

HendrikThePendric: > Inside that function, there's a condition that compares the *response bodies* and only returns true if they're equal.

BTW: the highlighted part is incorrect: `findMatchingRequestStub` does not look at matching _response_ bodies, it looks at matching _request_ bodies.

Mohammer5: Ok, got it! thanks, now I also know, what nonDeterministic means

Mohammer5 (approved):
Code looks good to me now!

PR #107, opened 2020-12-03
refactor: restructure install cmd & create plugins package

Mohammer5:
BREAKING CHANGE: This PR changes the install command's implementation


packages/cli/codemods/add-plugin.js
```diff
@@ -0,0 +1,70 @@
...
+                path.parentPath &&
+                path.parentPath.parentPath &&
+                path.parentPath.parentPath.value &&
+                path.parentPath.parentPath.value.left &&
+                path.parentPath.parentPath.value.left.object &&
+                path.parentPath.parentPath.value.left.property &&
+                path.parentPath.parentPath.value.left.object.name ===
+                    'module' &&
+                path.parentPath.parentPath.value.left.property.name ===
+                    'exports'
```
varl, line 45: Man, optional chaining (Node >=14) would be amazing here... :cry: 

 ```js
path.parentPath?.value?.left?.property?.name === 'module' && path.parentPath?.value?.left?.object?.name === 'exports'
```

File: packages/cli/src/commands/install.js [hunk 2]
```diff
@@ -1,83 +1,210 @@
...
+            state,
         })
-        .option('verbose', {
-            describe: 'Detailed output',
-            type: 'boolean',
-            default: false,
+        state = await setNonGlobalStepDefinitions({
+            options: providedOptions,
+            state,
         })
```
varl, line 146: I think this is easier to read than it was before :+1: . If we want to isolate the mutation of `state`, we could wrap it in a function that builds out the state which would allow us to at least protect the `state` variable from unintentional reassignment in the `handler` scope.

E.g.

```js
const buildState = async (oldState, options) => {
    let newState = { ...oldState }
    newState = await setBaseUrl({ options, newState })
    newState = await setRecordVideo({ options, newState })
    newState = await setProjectId({ options, newState })
    newState = await setLoginUser({ options, newState })
    newState = await setLoginPassword({ options, newState })
    newState = await setLoginBackendUrl({ options, newState })
    newState = await registerCommands({ options, newState })
    newState = await setDataTestPrefix({
        options,
        newState,
        paths,
    })
    newState = await usePluginChromeAllowXSiteCookies({
        options,
        newState,
    })
    newState = await usePluginCucumberPreprocessor({
        options,
        newState,
    })
    newState = await setTestFilesToFeatureFiles({
        options,
        newState,
    })
    newState = await setNonGlobalStepDefinitions({
        options,
        newState,
    })

    return newState
}

const state = await buildState(initialState, providedOptions)
```

A minor change with a slight benefit, and if we do this it might be tempting to extract the `buildState` function into its own file. On one hand it is nice to be able to see the construction of the state in the handler without having to dig around for it, on the other hand, it may be more clear what the business logic is if it is extracted.

packages/cli/src/commands/install/askForPackageManager.js
```diff
@@ -0,0 +1,17 @@
...
+module.exports.askForPackageManager = async () => {
+    const prompt = inquirer.createPromptModule()
+    const userInput = await prompt([
+        {
+            type: 'list',
+            name: 'packageManager',
+            message:
+                'This step will install the @dhis2/cypress-commands package. Choose your package manager',
+            choices: ['yarn', 'npm', 'skip'],
+            default: 0,
```
varl, line 12: Could the default option be determined based on the existence of a `yarn.lock` or a `package-lock.json` file?

If a project has neither, or both, we could present `skip` as default.

packages/cli/src/commands/install/mergeSupportFileImports.js
```diff
@@ -0,0 +1,24 @@
...
+const onlyUnique = (value, index, self) => self.indexOf(value) === index
+
+module.exports.mergeSupportFileImports = (support, paths) => {
+    const importNames = support.filter(onlyUnique)
+
+    const codemod = path.join(
+        __dirname,
+        '..',
+        '..',
+        '..',
```
varl, line 16: Could we simplify these by using a `paths.PROJECT_ROOT` constant instead of duplicating this lookup across all install controllers?

varl (commented):
Code looks nice! I'll post some comments now and approve after I play around with it on the CLI.

packages/cli/src/commands/install.js
Mohammer5: I was thinking about this as well and, at least right now, prefer having it in the install command as is.
There's an alternative though: We could add the handler function to the options in `packages/cli/src/commands/install/options.js` and the just use a function (e. g. `processOptions`). The outcome would be more or less what you're suggesting, but it'd have the advantage that the option and its handler are coupled together / co-located in the code

varl: I'm OK with it as it is now, and we can refactor down the line if we need/want to.

Mohammer5: Too late

packages/cli/src/commands/install/mergeSupportFileImports.js
Mohammer5: I used `CLI_TOOL_ROOT` and `PROJECT` is ambigious

packages/cli/src/commands/install/askForPackageManager.js
Mohammer5: Implemented

packages/cli/codemods/add-plugin.js
Mohammer5: yeah...

Mohammer5:
@varl le done

docs/developer.md
```diff
@@ -0,0 +1,5 @@
+# Deverloper docs
```
varl, line 1:
```suggestion
# Developer docs
```

docs/developer/install.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Handlers
+
+The handlers have to return the complete state. If the option is turned off (or
+other conditions inside the handler determine that the state should not be
+changed), these return the state that's been passed in.
+
+This is the function signature:
+
+```js
```
varl, line 104:
```suggestion
```ts
```

packages/cli/src/commands/install.js
```diff
@@ -1,83 +1,100 @@
...
+
+        await applyState({ state, packageManager, paths })
+
+        if (Object.values(state.cypressEnvJson).length) {
+            log.warn(
+                'Make sure to add `cypress.env.json` to the `.gitignore` file!'
+            )
         }
     } catch (e) {
+        console.error(e)
```
varl, line 91: We shouldn't need both `log.error`and `console.error`?

packages/cli/src/utils/fs.js
```diff
@@ -30,37 +30,44 @@ const addToJson = (filepath, data, overwrite = false) => {
...
+    }
+
+    const contents = fs.readFileSync(filePath, { encoding: 'utf8' })
+    return contents
 }
 
 const write = (filepath, data, overwrite) => {
     const exists = fs.existsSync(filepath)
-    const content = JSON.stringify(data, null, 4)
+    //const content = JSON.stringify(data, null, 4)
```
varl, line 55: Drop this line?

Mohammer5:
I've added developer docs for the install command, please have a look at that one as well

varl (approved):
Some minor comments, other than that looks good to go from my side. The install command seems to work nicely! :muscle:

PR #115, opened 2020-12-17
refactor(get find commands): rename to get-/findWithDataTest

Mohammer5:
Blocked by: dhis2/cli-utils-codemods#8
Blocked by: dhis2/cli#397
Fixes CLI-10

BREAKING CHANGE: This removes overwriting the existing get and find
commands. Instead new commands will be provided: `getWIthDataTest` and
`findWIthDataTest


HendrikThePendric (approved):
🎉

varl (approved):

HendrikThePendric (changes requested):
After some more consideration, I think we should tweak the command names somewhat:
```
getWithDataTest => getByDataTest
findWithDataTest => findByDataTest
```
I think this will feel more natural since devs are used to methods like `getElementById`, `getElementsByClassName`, etc.

Mohammer5:
@HendrikThePendric This idea has already been proposed by @varl at an earlier stage. I'm against it as we're not "getting" by a data test value, we're getting by a selector that contains the data-test selector syntax.

So if this was how it worked: `getByDataTest('button-icon')`, then the name would be ok
But as it's not working that way, it's not comparable to `getElementById` or `getElementByClassName` and is more like `document.querySelectorAll`.

`queryWithDataTest` does not work as we have to differentiate between `get` and `find`. Hence the names `getWithDataTest` and `findWithDataTest`

HendrikThePendric (commented):
I'm changing my review status to neutral. I don't fully understand the need for this so will leave it up to others to approve or request changes. From my perspective it would make more sense to simple deprecate the overwritten commands and move to native CSS selectors.

HendrikThePendric (approved):
OK, what I previously attempted to do isn't possible in GitHub and my review got stuck on "Requested changes". I've understood that this approach has been discussed with @varl so I'll approve as well.

PR #121, opened 2021-01-04
fix(commands): add `-W` to `yarn add` when in yarn workspaces root

HendrikThePendric:
The command `d2-utils-cypress install` would not complete successfully in a yarn workspaces repo due to the missing `-W` flag. This fixes it, but I wonder if there would be a more elegant way to do this...

### DIFF
--- a/packages/cli/src/commands/install/ensureDependency.js
+++ b/packages/cli/src/commands/install/ensureDependency.js
@@ -3,12 +3,18 @@ const exec = require('@dhis2/cli-helpers-engine').exec
 const log = require('@dhis2/cli-helpers-engine').reporter
 
 const isYarnWorkspacesRoot = () => {
-    const rootDir = process.cwd()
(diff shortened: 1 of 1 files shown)


amcgee (changes requested):
@HendrikThePendric I think this will throw if `package.json` doesn't exist or isn't valid JSON, so if going with this approach it at least needs a `try/catch`

HendrikThePendric:
I did consider that, but thought I would just let the error throw in any case. If there is no package.json or it contains invalid json, this script failing is the least of the repo's worries right?

amcgee:
It could fail just because it's being run in a directory that doesn't happen to be a package root - in that case it would be much nicer to say "install must be run from the root of an npm package, but no package.json was found" rather than simply an IO failure with a nasty stack trace.

amcgee:
Technically we should find-up to the nearest package root anyway, rather than assuming the CWD is the root

amcgee (approved):
LGTM - suggested a minor update to the message

PR #122, opened 2021-01-04
fix(commands): move jscodeshift to regular dependencies

HendrikThePendric:
jscodeshift is not a dev dependency as it is used in the install command


ismay (approved):
Yeah looks correct to me 👍

PR #124, opened 2021-01-05
fix(commands): adjust generated code changes

HendrikThePendric:
Two things were going wrong related to code generation:

1. `mergeSupportFileImports` was adding a default import of the plugins package to the support index file. This was incorrect and code that does correctly imports things into the support and plugins index files is already in place. As such, this function could simply be removed. 
2. The way we add plugins has changed and codemods had been introduced to insert code that accommodates that. However, we didn't have any code in place that would remove the legacy code and this was throwing errors. I've added a new codemod to remove the old stuff.


ismay:
It seems nice to have simple tests for each individual codemod, as an easy way to verify that it does what we expect. For this PR at least for the ones we're changing. Is that attainable with the current code structure?

HendrikThePendric:
I've added a test in 40fcd70 and also tweaked some related things:
- Running `yarn test` from the root will now execute the tests from both the `cli` and the `cypress-commands` package concurrently
- This command is also added as a husky pre-commit hook

Strictly speaking this is out-of-scope for this PR but it wasn't much work and aligns this repo's setup a bit better with our others.

Also I noticed there was already a codemod test present here:
`packages/cypress-commands/codemods/__test__/replace-get-find-with-xxxxGetWithDataTest.test.js`
However, this test is completely manual. I think at some point we should update this test to use the helpers I've used and are documented [here](https://github.com/facebook/jscodeshift/blob/master/README.md#unit-testing). I'll create a JIRA issue for this.

packages/cli/codemods/__tests_/purge-plugin-legacy-code.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+defineInlineTest(
+    transform,
+    transformOptions,
+    'plugins(on, config)',
+    '',
+    'Removes the legacy function call'
+)
+
+defineInlineTest(
```
ismay, line 21: Maybe for readability using variables for the input and output would improve the formatting. Looks kind of off with the template string not matching the surrounding indent.

ismay:
> I've added a test in [40fcd70](https://github.com/dhis2/cli-utils-cypress/commit/40fcd7017c715e544130655e79629d60c6363d1f) and also tweaked some related things:
> 
>     * Running `yarn test` from the root will now execute the tests from both the `cli` and the `cypress-commands` package concurrently
> 
>     * This command is also added as a husky pre-commit hook
> 
> 
> Strictly speaking this is out-of-scope for this PR but it wasn't much work and aligns this repo's setup a bit better with our others.
> 
> Also I noticed there was already a codemod test present here:
> `packages/cypress-commands/codemods/__test__/replace-get-find-with-xxxxGetWithDataTest.test.js`
> However, this test is completely manual. I think at some point we should update this test to use the helpers I've used and are documented [here](https://github.com/facebook/jscodeshift/blob/master/README.md#unit-testing). I'll create a JIRA issue for this.

Sounds good 👍

ismay (approved):
Changes look good to me. Only thing that stood out to me is the left alignment of the input and output not matching the surrounding indent for the tests. I understand that that's a consequence of the template literals, but maybe readability could be improved there by using variables for the inputs and outputs? That's very minor so I'll leave that to what you prefer @HendrikThePendric.

PR #131, opened 2021-01-27
fix: address lifecycle and response type bugs (CLI-25)

HendrikThePendric:
Christos tried using the Network Shim in the capture app. Two issues were uncovered:

1. We were creating an alias in the before hook which wasn't available in the beforeEach hook. Not sure why this was working in the example app, but according to the cypress docs all aliases are cleared between tests, so we need to create the aliases in beforeEach and afterEach.
2. We encountered an endpoint that returned an unquoted number and cy.intercept didn't like this. So anything that's not a string or an object needs to be stringified.

This PR fixes them both.


packages/cypress-commands/src/setups/enableNetworkShim/index.js
```diff
@@ -38,7 +39,7 @@ export function enableNetworkShim({
         })
     })
 
-    after(() => {
+    afterEach(() => {
```
amcgee, line 42: Won't this change cause the file to be written after each individual test?  Is that a problem?

amcgee: My understanding is that this will overwrite the file with a bit more information after every test, which is probably OK, but could cause performance or even race condition problems when files are large...?

amcgee (commented):
Haven't tested but code looks good!  Just one question, probably caused by my outdated mental model

packages/cypress-commands/src/setups/enableNetworkShim/index.js
HendrikThePendric: Yeah very true. However, since the cypress [clears aliases between tests](https://github.com/cypress-io/cypress/issues/665#issuecomment-330600042), we really do have to stick to the `beforeEach` and `afterEach` hook. This approach is far from ideal, but luckily the cypress-plugin based approach which we'll introduce soon solves this issue.

I've already tested this current implementation in the data-capture app and found the following:
- No race conditions with reading/writing files occurred (could have been a stroke of luck, but it was quite a substantial test-suite)
- I compared the cumulative/total run-time of a capture-run with a normal run and found that the capture run was 17% slower. Quite bad, but not terrible.

My recommendation for now would be to merge the current PR so we have a fix for v5, but to also take this as a cue to press ahead with the plugin based version, which will be in v6.

HendrikThePendric: FYI, this will be the implementation when we switch to the cypress-plugin based approach. This should be much less problematic because the state object is created by the plugin when the run starts and files are written when the run ends, during the run it's simply getting and setting this state:
https://github.com/dhis2/cli-utils-cypress/blob/b8b67b2affaf114a4777d4d90f79dc62a048e5b8/packages/cypress-commands/src/setups/enableNetworkShim/index.js#L6-L41

amcgee: Great, thanks for the analysis!  I agree, let's get this in

amcgee (approved):

PR #146, opened 2021-02-23
fix(install): add imports to support file

Mohammer5:
Fixes CLI-29 ([link](https://jira.dhis2.org/browse/CLI-29))

Fixes a regression created by #124


HendrikThePendric (approved):
Tested this locally and it's working as expected.

PR #176, opened 2021-04-22
fix(network-shim): ensure in-test fixtures are used instead of shim fixtures

HendrikThePendric:
The initial issue was fixed by upgrading cypress to v7, which includes the correct overriding behaviour for `cy.intercept`.
Due to the major version bump there was a regression in both capture and stub mode: no requests were being intercepted by the network-shim at all. This was fixed by passing regexes to `cy.intercept` instead of strings.

A new test was introduced to the network-shim example-app to assert the in-test fixtures take precedence over the network-shim fixtures.

Once capturing was working properly again, we encountered another regression due the major version bump: stub mode was producing network errors. After investigation it turned out the network errors were happening because of a `content-encoding` property in the header. This is a cypress bug which I filed here: https://github.com/cypress-io/cypress/issues/16420. This bug actually isn't that relevant for the network-shim because want to filter response and request headers properties anyway. I have implemented an allow-list mechanism and omitted `content-encoding`.

Interestingly, we didn't encounter this cypress `content-encoding` headers until we tried the network-shim in the sms-configuration-app. The `testing-network-shim-app` example app was unaffected. This was a reason to introduce a new DHIS2 platform example app. This new app also includes a cypress test, which functions as a smoke test for the functioning of the network-shim in a DHIS2 platform app.

And finally, since it has become painfully clear that cypress changes a lot and that these changes might cause the network-shim to break, @varl and I have tweaked the GitHub workflows so that an e2e suite runs on CI. This should guard against regression. More details about the test have been added to the docs.

BREAKING CHANGE: bumps cypress 1 major version, to v7


HendrikThePendric:
@Mohammer5 thanks for the additional commit. Would you like to request changes? If not, please approve so we can release this fix.

Mohammer5:
I tested it with the sms config app and it didn't work "plug'n'play" style. Capturing seems to have worked, but during stub mode, the tests did not login successfullly:

image

I had to make some adjustments to the sms config app to work with this version of the cli utils as they use cypress@7 and the sms config app still used v5 of our tool, the branch is: Prepare_network_shim

examples/platform-app/package.json
```diff
@@ -0,0 +1,35 @@
...
+        "lint": "yarn d2-style check",
+        "lint:staged": "yarn lint --staged",
+        "format": "yarn d2-style apply",
+        "format:staged": "yarn format --staged",
+        "cy:start": "BROWSER=none REACT_APP_DHIS2_BASE_URL=http://localhost:8080 REACT_APP_DHIS2_API_VERSION=37 yarn start",
+        "cy:open": "d2-utils-cypress open --appStart 'yarn cy:start'",
+        "cy:run": "d2-utils-cypress run --appStart 'yarn cy:start'",
+        "cy:capture": "d2-utils-cypress run --capture --appStart 'yarn cy:start' --serverMinorVersion 37",
+        "cy:stub": "d2-utils-cypress run --stub --appStart 'yarn cy:start' --serverMinorVersion 37",
+        "cy:stub-debug": "DEBUG=cypress:network:*,cypress:net-stubbing* d2-utils-cypress run --stub --appStart 'yarn cy:start' --serverMinorVersion 37"
```
Mohammer5, line 20: Not sure if this should make it into the release, but it's not crucial functionality and not blocking any other script, so I don't have a strong opinion on this

Mohammer5 (changes requested):
The only real change requests I have are what we already discussed:

### Please add docs about the quirk with testing the `testing-network-shim-app` example app:

```
npx --no-install cypress cache clear
find ./ -iname node_modules -type d | xargs rm -r 
yarn install
```

### Add docs or a script to test the whole network shim in both example apps

Mohammer5 (approved):

PR #201, opened 2021-05-20
fix(network-shim): fix before hook bug

HendrikThePendric:
These changes were created because we encountered a problem when trying to implement the NetworkShim in the sms-configuration-app. It seemed that in some cases the network shim intercept callback was still being executed even if there was an intercept with fixture declared in the test itself. And because the NetworkShim was throwing an error the test failed.

I have spent a long time trying to write a reproduction case in the platform-app example. After trying for quite long time and still being unable to reproduce the exact problem I discussed things with @varl. Because the issue could be reproduced consistently in the sms-configuration-app and also solved there fairly easily, and because this PR will get merged to `alpha` first, it was OK to just write a fix without a reproduction case.

So at the core of the current fix lies the notion that encountering a missing fixture during a stub run _could be_ a bad thing, but it doesn't _have to be_ a problem since a test could have declared a `cy.intercept()` with a fixture or static response object too. So instead of throwing an error I collect the missing fixtures and once the run is done I will show a warning. I only show this warning when:
- There are missing fixtures
- We are in stub mode
- The test has 1 or more failures (because if the suite passed the missing fixtures obviously weren’t a problem)

Besides changes related directly to the problem described above, I have also ported the test suite of the example platform-app to use a cypress-cucumber test-suite, so now we have:
- One CRA example app with vanilla cypress test suite
- Another example app, which is a DHIS2 Platform App with a cypress-cucumber test suite

I think we can these changes in too, even though they aren't strictly related to the issue at hand.


HendrikThePendric:
UPDATE: I tried the current fix in the sms-configuration-app and everything seems to be working correctly.

packages/cypress-commands/src/setups/enableNetworkShim/stubRequests.js
```diff
@@ -47,19 +47,28 @@ function stubRequest(state, request) {
...
 
-    const responseBody = getRequesStubResponseBody(requestStub)
+        /*
+         * Reply with an dummy response to avoid the request
+         * from being sent to the destination server
+         */
+        request.reply({
+            body: {},
+            headers: {},
+            statusCode: 200,
```
Mohammer5, line 59: I think 404 is the better status code here, as it's not found.
If there's a manual mock, then this should be irrelevant, right?

packages/cypress-commands/src/setups/enableNetworkShim/stubRequests.js
```diff
@@ -75,3 +84,28 @@ function getRequesStubResponseBody({
...
+    const isDuplicate = state.missingRequestStubs.some(
+        stub =>
+            stub.method === method &&
+            stub.path === path &&
+            stub.testName === testName
+    )
+
+    if (!isDuplicate) {
+        state.missingRequestStubs.push({ method, path, testName })
+    }
```
Mohammer5, line 110: I'm wondering if it might be worth to add the amount the duplicate has been encountered, so you also know how many times it went wrong

Mohammer5 (approved):
Just two really minor things that are not that important, so you can take it or leave it, approved :+1:  Looks very good to me!

packages/cypress-commands/src/setups/enableNetworkShim/stubRequests.js
HendrikThePendric: Great suggestion!

packages/cypress-commands/src/setups/enableNetworkShim/stubRequests.js
HendrikThePendric: Deferred for now

packages/cypress-plugins/src/plugins/networkShim/reportMissingRequestStubs.js
```diff
@@ -0,0 +1,23 @@
+const log = require('@dhis2/cli-helpers-engine').reporter
+
+module.exports = function reportMissingRequestStubs(state, results) {
+    if (
+        results.totalFailed > 0 &&
+        Array.isArray(state.missingRequestStubs) &&
+        state.missingRequestStubs.length > 1
```
amcgee, line 7: Might be missing something, but should this be >=1 instead of >1 ?

amcgee (commented):
Looks good!  Just one comment, might be intentional?

packages/cypress-plugins/src/plugins/networkShim/reportMissingRequestStubs.js
HendrikThePendric: Shoot you are right, `> 0` would have made most sense. I'll prepare a fix PR and will merge that in cowboy style. Yeehaw 🤠 . (alpha is not protected)

amcgee: Could that be related to not being able to reproduce..?

PR #215, opened 2021-05-31
refactor(cli): remove npx

varl:
This removes the use of `npx` to run commands, instead using the functions from `cli-helpers-engine` to run commands.

Since these engine functions spawn processes, we cannot use `concurrently` programatically to run them as it only accepts command strings that it then spawns internally.

We also cannot use `npx` to run `wait-on` the same way as we did before either as it pulled it implicitly from `npm` with `npx`. 3f2469b uses wait-on programmatically to wait 30s for the cypress.json defined `baseUrl` to become available.


varl:
This depends on: https://github.com/dhis2/cli-helpers-engine/pull/96

Mohammer5:
This PR does more than removing npx, it also introduces a breaking change: it removes the internal usage of `concurrently` as well as `wait-on`. I can see arguments for removing it (as well as for keeping it). Could you explain what made you remove them @varl?

varl:
`wait-on` is added back in a separate commit, but `concurrently` cannot be used internally when we remove `npx`. And when we remove `concurrently`, that means the `--app-start` flag will be difficult to implement.

Lifting `concurrently` "up" one level allows the consumer to define e.g. `concurrently 'yarn start' 'd2-utils-cypress run'` in a script.

Since we load the `baseUrl` from the `cypress.json` configuration instead of assuming that it's [`localhost:3000`](https://github.com/dhis2/cli-utils-cypress/blob/master/packages/cli/src/commands/common/sharedCLIOptions.js#L32-L36) we can drop that `--waitOn` flag as well.

True, we wait for a `HEAD 200` instead of a `GET 200`, but if we need to we can adjust the prefix programmatically. I haven't seen a need to do so, so I have left it out.

varl:
There are other changes that we should do here, to make it closer aligned between local runs and CI, but I have chosen to leave those out of this PR.

We want to do that because at least 3 developers have spent days figuring out how d2-utils-cypress / Cypress Action / network-shim / environment variables interact, so it needs to behave as consistently as possible across local envs and CI.

This does more than remove `npx`, true, and leads to the breaking changes. `npx` is a fundamental piece of why this software works the way it does, so removing it forces re-evaluation of how it operates internally.

The rationale for the breaking changes was this:

Since there are already breaking changes in the `alpha` (to where this is slated), and these changes are simplifications (removals), and re-implementing the fundamental parts (running, and managing, spawned processes concurrently) is complicated and, as I see it, out of scope, this is a good trade-off for an `alpha` version.

Mohammer5:
I fail to see why `npx` is needed for `concurrently` as it's used to start either `cypress` or `cypress-tags`.

Right now, I assume that you've encountered an issue with using `concurrently` with `cypress`/`cypress-tag` directly, hence the use of `npx` (I think I am the one who initially added it, but I can't recall why I did that tbh)

varl:
> I fail to see why `npx` is needed for `concurrently` as it's used to start either `cypress` or `cypress-tags`.
> 
> Right now, I assume that you've encountered an issue with using `concurrently` with `cypress`/`cypress-tag` directly, hence the use of `npx` (I think I am the one who initially added it, but I can't recall why I did that tbh)

With `npx` what we do is build a string e.g. `npx --no-install wait-on 'http://...' && npx --no-install cypress run ...` and pass that string to `concurrently`. Then the concurrently executes the actual command.

If we remove `npx`, we cannot pass in `wait-on 'http://...' && cypress run ...` directly to concurrently because it will not be able to resolve to the correct `node_modules/.bin`. To workaround that we use the `cli-helpers-engine`'s `bin` function, that can resolve to the correct `node_modules/.bin`, but that has a caveat of actually spawning a process for the command and running it directly instead of generating a command string that `concurrently` uses.

Does that explain it?

examples/platform-app/package.json
```diff
@@ -8,26 +8,17 @@
...
         "format:staged": "yarn format --staged",
         "cy:start": "BROWSER=none yarn start",
-        "cy:open": "d2-utils-cypress open --appStart 'yarn cy:start'",
-        "cy:run": "d2-utils-cypress run --appStart 'yarn cy:start'",
-        "cy:capture": "d2-utils-cypress run --capture --appStart 'yarn cy:start' --serverMinorVersion 37",
-        "cy:stub": "d2-utils-cypress run --stub --appStart 'yarn cy:start' --serverMinorVersion 37",
+        "cy:open": "concurrently 'yarn:cy:start' 'd2-utils-cypress open' --kill-others",
+        "cy:run": "concurrently 'yarn:cy:start' 'd2-utils-cypress run' --kill-others",
+        "cy:capture": "concurrently 'yarn:cy:start' 'd2-utils-cypress run --capture --serverMinorVersion 37' --kill-others",
+        "cy:stub": "concurrently 'yarn:cy:start' 'd2-utils-cypress run --stub --serverMinorVersion 37' --kill-others",
```
Mohammer5, line 19: I think `concurrently` should be a devDependency of this example app and not be specified in the root package.json as that decouples the dependency from the workspace that uses it (unlike d2-style for example, which operates on the root level)

examples/testing-network-shim-app/package.json
```diff
@@ -30,10 +22,10 @@
...
         "cy:start": "BROWSER=none yarn start",
         "cy:serve-and-start": "concurrently \"yarn serve\" \"yarn cy:start\"",
-        "cy:open": "d2-utils-cypress open --appStart 'yarn start'",
-        "cy:run": "d2-utils-cypress run --appStart 'yarn start'",
-        "cy:capture": "d2-utils-cypress run --capture --appStart 'yarn cy:serve-and-start' --serverMinorVersion 36",
-        "cy:stub": "d2-utils-cypress run --stub --appStart 'yarn cy:start' --serverMinorVersion 36",
+        "cy:open": "d2-utils-cypress open",
+        "cy:run": "d2-utils-cypress run",
+        "cy:capture": "concurrently 'yarn cy:serve-and-start' 'd2-utils-cypress run --capture --serverMinorVersion 36'",
+        "cy:stub": "concurrently 'yarn serve' 'yarn cy:start' 'd2-utils-cypress run --stub --serverMinorVersion 36'",
```
Mohammer5, line 28: Same as above

package.json
```diff
@@ -39,11 +34,14 @@
...
     "devDependencies": {
-        "@dhis2/cli-style": "^7.2.1",
-        "@dhis2/cli-utils-docsite": "^1.3.0",
+        "@babel/core": "^7.12.3",
+        "@babel/preset-env": "^7.12.1",
         "@dhis2/cli-app-scripts": "^4.0.4",
-        "jest": "^26.6.3",
+        "@dhis2/cli-style": "^8.3.0",
+        "@dhis2/cli-utils-docsite": "^1.3.0",
+        "concurrently": "^6.2.0",
```
Mohammer5, line 42: See above

Mohammer5 (approved):
"Requested" some really minor changes, it's not that important, so I approve and leave the decision up to you

examples/platform-app/package.json
varl: Will manage this in the follow-up PR. :+1:

PR #217, opened 2021-06-01
fix: use upper-snake-case environment variable names

HendrikThePendric:
BREAKING CHANGE: all environment variables are now in `UPPER_SNAKE_CASE`, aligned with the bash convention. This means apps  will have to change their `cypress.env.json` and perhaps some scripts in `package.json`

### DIFF
--- a/docs/commands/login.md
+++ b/docs/commands/login.md
@@ -3,9 +3,9 @@
 This command will log a specified user in through the login endpoint.
 It uses the following required environment variables:
 
--   `dhis2_username`
--   `dhis2_password`
--   `dhis2_base_url`
+-   `DHIS2_USERNAME`
+-   `DHIS2_PASSWORD`
+-   `DHIS2_BASE_URL`
 
 Please refer to the [official
 docs](https://docs.cypress.io/guides/guides/environment-variables.html) for
--- a/docs/guides/add-login-credentials.md
+++ b/docs/guides/add-login-credentials.md
@@ -2,9 +2,9 @@
 
 The cypress utility expects three values to be present in that file:
 
-1. `dhis2_base_url`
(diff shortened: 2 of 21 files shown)


varl (approved):
I agree profusely with adhering to this long-standing practice for environment variables names.

ismay (commented):
Should we not add the `BREAKING CHANGE` label in the commit body to make sure it ends up in the changelog? Otherwise people might miss this.

varl:
@ismay good catch.

Mohammer5:
I assumed that would be applied automatically when merging the PR, which should copy the PR's description into the merge's commit message body field?

varl (changes requested):
I've reconsidered a bit when I was working on #218. This is slightly more nuanced than I initially thought.

There is value in keeping the operating system environment variables separate from the cypress environment variables.

See #218 to see what I mean, and refer to the Cypress docs[1] for why they support both: 

> In Cypress, "environment variables" are variables that are accessible via Cypress.env. These are not the same as OS-level environment variables. However, it is possible to set Cypress environment variables from OS-level environment variables.

[1] https://docs.cypress.io/guides/guides/environment-variables

ismay:
Yeah the reason I asked for this is because it gets confusing when you're setting cypress env vars from the OS. Which we're doing on CI. If we keep the existing approach we'll have both lower- and uppercased env vars on CI (and they'll be meaningful, cypress makes a distinction between upper and lowercased env vars).

If it were only cypress env vars I'd agree, but since we're crossing over into OS env vars I'd vote for uppercasing.

HendrikThePendric:
Closing in favour of https://github.com/dhis2/cli-utils-cypress/pull/217

PR #218, opened 2021-06-02
refactor: simplify the network-shim and d2-utils-cypress

varl:
> See this commit for an explanation: https://github.com/dhis2/cli-utils-cypress/commit/90827973cbdaf841b951b7d29ebd501d574c4da5:

Refactor the way the way d2-utils-cypress relates to cypress, and how it
is intended to be used in a project.

The biggest change is that a consumer does not use d2-utils-cypress to
set up the environment for Cypress and then automatically run Cypress
with those changes.

Instead, the consumer runs Cypress directly.

    cypress run

After initial configuration in either cypress.json through the "env"
key, or in cypress.env.json, the user can activate the plugin through:

    cypress run --env network=capture
    cypress run --env network=stub
    cypress run --env network=live

"live" is the default if the network env property is not set, so it can
be omitted, but for clarity it is good practice to add it to the
cypress.env.json:

    {
        "dhis2_username": "admin",
        "dhis2_password": "district",
        "dhis2_base_url": "http://localhost:8080",
        "dhis2_server_api_version": "37"
        "dhis2_datatest_prefix": "networkshim-platformapp",
        "network": "live",
    }

The cypress-plugin uses the env var "dhis2_server_api_version" to switch
the expected API version of the DHIS2 server. A default can be supplied
in a config file as above, or overridden per run. For example a stub run
against version 36:

    cypress run --env network=stub,dhis2_server_api_version=36

BREAKING CHANGE: Drop run and open commands
We want to be consistent with how Cypress runs locally and in CI and
since we cannot use d2-utils-cypress in CI, we shouldn't run it through
d2-utils-cypress locally either.

BREAKING CHANGE: dhis2_server_minor_version renamed to
dhis2_server_api_version

BREAKING CHANGE: dhis2_api_stub_mode renamed to network
Instead of describing the mode of the plugin, it is a bit easier to
understand if we speak in terms of the network:
- do we want to capture the network traffic (network=capture),
- do we want to stub it (network=stub),
- or do we want to run it against a live backend (network=live)?

BREAKING CHANGE: 'DISABLED' renamed to 'LIVE'
To better describe the state of the network when running tests instead
of describing the state of the plugin, DISABLED is now LIVE.

BREAKING CHANGE: isDisabledMode renamed to isLiveMode.
Similar to the above, to better describe the state of the network vs.
the state of the plugin, replace usages of isDisabledMode with
isLiveMode.

BREAKING CHANGE: 'CAPTURE'|'STUB'|'LIVE' are now lowercase when passed
to the environment.
Replace network=LIVE with network=live.


Mohammer5:
This seems to be still WIP (you're still adding commits)? Just let us know when you think it's ready for review

varl:
> This seems to be still WIP (you're still adding commits)? Just let us know when you think it's ready for review

Yeah, I'll request reviews when it's "done".

varl:
OK, this works now.

examples/platform-app/cypress.env.json
```diff
@@ -2,5 +2,7 @@
     "dhis2_username": "admin",
     "dhis2_password": "district",
     "dhis2_base_url": "http://localhost:8080",
-    "dhis2_datatest_prefix": "networkshim-platformapp"
+    "dhis2_datatest_prefix": "networkshim-platformapp",
+    "network": "live",
+    "dhis2_api_version": "37"
```
varl, line 7: Names can be bikeshedded, but I chose these because it's reasonably clear that we are talking about the dhis2 core api version here, and that isn't really a "minor", and shorter is a good thing, imo.

We may also want to extend the shim to handle e.g. `apphub_api_version` and I think that's fine. Not everything needs to be prefixed with dhis2.

I chose `network` because it's declarative and short, e.g. `network=capture`, `network=stub`, and `network=live` is pretty clear to me in terms of what they do with the network.

examples/platform-app/package.json
```diff
@@ -13,10 +13,10 @@
...
         "format:staged": "yarn format --staged",
         "cy:start": "BROWSER=none yarn start",
-        "cy:open": "concurrently 'yarn:cy:start' 'd2-utils-cypress open' --kill-others",
-        "cy:run": "concurrently 'yarn:cy:start' 'd2-utils-cypress run' --kill-others",
-        "cy:capture": "concurrently 'yarn:cy:start' 'd2-utils-cypress run --capture --serverMinorVersion 37' --kill-others",
-        "cy:stub": "concurrently 'yarn:cy:start' 'd2-utils-cypress run --stub --serverMinorVersion 37' --kill-others",
+        "cy:open": "wait-on 'http://localhost:3000' && cypress open",
+        "cy:run": "wait-on 'http://localhost:3000' && cypress run",
+        "cy:capture": "concurrently 'yarn cy:start' 'yarn cy:run --env network=capture' --kill-others --success first",
+        "cy:stub": "concurrently 'yarn cy:start' 'yarn cy:run --env network=stub' --kill-others --success first",
```
varl, line 19: `wait-on` is moved to the outside, and `concurrently` only exists on the outside (since the `npx` removal). It's a bit more verbose, but this can be extracted to a script in an app for clarity and brevity.

The `d2-utils-cypress install` command can be extended to set that up for us.

examples/testing-network-shim-app/.eslintignore
```diff
@@ -0,0 +1 @@
+*
```
varl, line 1: I ran into some annoying problems with ESLint rules in this App so I disabled them.

examples/testing-network-shim-app/.env
```diff
@@ -1 +1,3 @@
-SKIP_PREFLIGHT_CHECK=true
\ No newline at end of file
+SKIP_PREFLIGHT_CHECK=true
+DISABLE_ESLINT_PLUGIN=true
+EXTEND_ESLINT=true
```
varl, line 3: `react-scripts` wasn't starting, so overriding their ESLint things to allow the App to start even if there are violations. It's an example app, so as long as the tests pass I don't really care about if there are lint errors...

examples/testing-network-shim-app/cypress.env.json
```diff
@@ -2,5 +2,7 @@
     "dhis2_base_url": "http://localhost:1337",
     "dhis2_username": "admin",
     "dhis2_password": "district",
-    "dhis2_datatest_prefix": "testing-network-shim-app"
+    "dhis2_datatest_prefix": "testing-network-shim-app",
+    "dhis2_api_version": "36",
+    "network": "live"
```
varl, line 7: `network` defaults to `live`, but added it since it's an example env config.

The api version can be set to a default as well for less verbose `--env` flags. We could generate different `cypress.env.json` files for different environments with different versions to run against too, if we don't add the `cypress.env.json` to Git.

varl:
@ismay With this change, the workflow in the scheduler-app branch `update-cypress` could look like:

```diff
               uses: cypress-io/github-action@v2
               with:
                   start: yarn cypress:start
-                  command: yarn d2-utils-cypress run --stub --serverMinorVersion 37 --appStart ''
+                  command: cypress run --env network=stub,dhis2_api_version=37
                   wait-on: 'http://localhost:3000'
                   wait-on-timeout: 300
               env:
-                  cypress_dhis2_base_url: https://debug.dhis2.org/dev
-                  cypress_record_key: ${{ secrets.CYPRESS_RECORD_KEY }}
+                  CYPRESS_DHIS2_BASE_URL: https://debug.dhis2.org/dev
+                  CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

Or

```diff
               uses: cypress-io/github-action@v2
               with:
                   start: yarn cypress:start
-                  command: yarn d2-utils-cypress run --stub --serverMinorVersion 37 --appStart ''
                   wait-on: 'http://localhost:3000'
                   wait-on-timeout: 300
+                  env: dhis2_base_url=https://debug.dhis2.org/dev,dhis2_api_version=37,network=stub
               env:
-                  cypress_dhis2_base_url: https://debug.dhis2.org/dev
-                  cypress_record_key: ${{ secrets.CYPRESS_RECORD_KEY }}
+                  CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

Or

```diff
               uses: cypress-io/github-action@v2
               with:
                   start: yarn cypress:start
-                  command: yarn d2-utils-cypress run --stub --serverMinorVersion 37 --appStart ''
                   wait-on: 'http://localhost:3000'
                   wait-on-timeout: 300
               env:
-                  cypress_dhis2_base_url: https://debug.dhis2.org/dev
-                  cypress_record_key: ${{ secrets.CYPRESS_RECORD_KEY }}
+                  CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                  CYPRESS_dhis2_base_url=https://debug.dhis2.org/dev
+                  CYPRESS_dhis2_api_version=37
+                  CYPRESS_network=stub
```

I think the last one with capital `CYPRESS_` and lowercase internal variable makes sense as per the docs here: https://github.com/cypress-io/github-action#env

It's a bit of a learned pattern but the `CYPRESS_` makes it clear that it's for Cypress. The `CYPRESS_RECORD_KEY` is different from those so that separation is also OK to me.

examples/platform-app/cypress.env.json
HendrikThePendric: I'm on board with these new names.

examples/testing-network-shim-app/.eslintignore
HendrikThePendric: See https://github.com/dhis2/cli-utils-cypress/pull/218/commits/5bc2da95146dbea10c1cdbf2751af3ded8ab4743 and b0abf9b. I re-enabled the linting and fixed the annoying lint errors. Hope this is OK with you.

examples/testing-network-shim-app/cypress/support/index.js
```diff
@@ -2,7 +2,7 @@ import { enableNetworkShim } from '@dhis2/cypress-commands'
 import { resetDb } from '../../src/json-server/resetDb'
 
 afterEach(() => {
-    if (Cypress.env('dhis2_api_stub_mode') === 'CAPTURE') {
+    if (Cypress.env('network') === 'capture') {
```
Mohammer5, line 5: I'm wondering if it'd make sense to use constants for these and expose them from the plugin package and use them here (as well as in the commands package)

package.json
```diff
@@ -36,12 +36,13 @@
...
-        "@dhis2/cli-app-scripts": "^4.0.4",
+        "@dhis2/cli-app-scripts": "^6.2.0",
         "@dhis2/cli-style": "^8.3.0",
         "@dhis2/cli-utils-docsite": "^1.3.0",
         "concurrently": "^6.2.0",
         "eslint-plugin-cypress": "^2.11.2",
         "eslint-plugin-prettier": "^3.2.0",
-        "jest": "^26.6.3"
+        "jest": "^26.6.3",
+        "wait-on": "^5.3.0"
```
Mohammer5, line 46: Do you want to move `wait-on` and `concurrently` into the example/testing apps in this PR or in another one?

packages/cypress-commands/src/setups/enableAutoLogin.js
```diff
@@ -1,5 +1,5 @@
 export const enableAutoLogin = () => {
-    if (Cypress.env('dhis2_api_stub_mode') === 'STUB') {
+    if (Cypress.env('network') === 'stub') {
```
Mohammer5, line 2: See other comment about constants

packages/cypress-plugins/src/plugins/networkShim/clearNetworkFixtures.js
```diff
@@ -5,7 +5,7 @@ const rimraf = require('rimraf')
 
 module.exports = function clearNetworkFixtures({ fixturesFolder, env }) {
     try {
-        const serverMinorVersion = env.dhis2_server_minor_version.toString()
+        const serverMinorVersion = env.dhis2_api_version.toString()
```
Mohammer5, line 8: This could be made a constant or symbol as well

examples/testing-network-shim-app/cypress/support/index.js
Mohammer5: Actually, the helper functions could be exposed instead

ismay:
Two other things that we should keep in mind as we're using the term "base url" in a couple places:

Screenshot 2021-06-02 at 17 03 24

We should make sure that we don't end up with confusing variable names as the term "base url" in cypress already has meaning. Also, they mention specifically not having to use an env var for setting this, but instead to use a config (unless you want to override it).

package.json
varl: It's moved in this PR, but in a monorepo the devDependencies go in the root package, not in the subpackages. I realize that it is slightly awkward for an example-app to omit some dependencies but having them here is more problematic than not having them.

varl:
> I've looked for explanations in the docs and issues on why cypress consistently has lowercased env vars in their examples and found nothing. So I'd just go with the more common approach of uppercasing env vars, as I see no compelling reason to do otherwise.

The first paragraph in the notice box here briefly talks about it: https://docs.cypress.io/guides/guides/environment-variables

There is a bit more here:

- https://docs.cypress.io/guides/references/configuration#Plugins
- https://docs.cypress.io/guides/references/configuration#Environment-Variables

They never go into detail why it is important to keep the Cypress env separate from the OS-level env, but I think it's a good distinction:

> In Cypress, "environment variables" are variables that are accessible via Cypress.env. These are not the same as OS-level environment variables. However, it is possible to set Cypress environment variables from OS-level environment variables.

The key is this:

- if you set `FOOBAR=true` on the OS-level, this is **not** visible to Cypress using `Cypress.env('FOOBAR')`.
- if you set `CYPRESS_foobar=true`, you can access it with `Cypress.env('foobar')`.
- In a plugin you can read `FOOBAR` from `process.env` and set it to the Cypress env to make it accessible with `Cypress.env('FOOBAR')` (in fact we do so with the `dhis2_*` env variables).

They could have expanded more on why the distinction is important in the docs, but I think it wise that the test code accesses the controlled and normalized Cypress environment through the global `Cypress.env()` function instead of allowing test code to use `process.env` directly through tests.

varl:
> We should make sure that we don't end up with confusing variable names as the term "base url" in cypress already has meaning. Also, they mention specifically not having to use an env var for setting this, but instead to use the config (unless you want to override it).

Base URL has a very specific meaning. It defines the URL to which end-point paths are appended to. This is the same reason that Cypress uses the term to identify the base URL of the site under test, so the consistency here is a good thing.

We could make it a bit more explicit to make the relationship with the `dhis2_api_version` very clear, and name it `dhis2_api_base_url`.

The reason we want to use the term base URL is because we are not specifying the end-point path to the API, which would be `https://debug.dhis2.org/dev/api`.

I haven't checked, but if we want to use the API URL instead, we could name the config key `dhis2_api_url` as it is clear that we should input the url to the API and not the base URL.

examples/testing-network-shim-app/.eslintignore
varl: Thanks !

varl:
@ismay I wrote a long comment about the differences and hid it as off-topic because you are right. It should not matter if we use `CYPRESS_dhis2_base_url` or `CYPRESS_DHIS2_BASE_URL` in our workflows. 

But it does matter. :thinking: 

`CYPRESS_DHIS2_BASE_URL` doesn't work with our plugin, but `CYPRESS_dhis2_base_url` does, similarly, `CYPRESS_dhis2BaseUrl` also does not work, but should because of the normalization Cypress does.

I think you uncovered a bug here. I'm looking into it.

varl:
Since Cypress considers our env variables different from their own, I'm leaning towards uppercase ones as well.

varl:
With the c3ea26b applied configuration now looks like:

```diff
               uses: cypress-io/github-action@v2
               with:
                   start: yarn cypress:start
-                  command: yarn d2-utils-cypress run --stub --serverMinorVersion 37 --appStart ''
                   wait-on: 'http://localhost:3000'
                   wait-on-timeout: 300
               env:
-                  cypress_dhis2_base_url: https://debug.dhis2.org/dev
-                  cypress_record_key: ${{ secrets.CYPRESS_RECORD_KEY }}
+                  CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                  CYPRESS_DHIS2_BASE_URL=https://debug.dhis2.org/dev
+                  CYPRESS_DHIS2_API_VERSION=37
+                  CYPRESS_DHIS2_PLUGIN_NETWORK_MODE=stub
```

examples/platform-app/cypress.env.json
varl: I may have jumped over board. :|

HendrikThePendric (approved):
🚀  LGTM

ismay:
@varl and I talked a bit on slack about the env var casing, just to speed up the discussion. So the core takeaways for me were:

- cypress.env is used to make certain env vars available in the _browser_ (so for code running in node you can still rely on process.env). I was under the impression that cypress also recommended using cypress.env for node code, that was a source of confusion for me, as node code can interact directly with shell env vars.
- it seems impossible to satisfy both the "env vars are global and global vars should be uppercased" rule for shell env vars, and satisfy the style of env vars cypress uses in their examples (lowercase, either snake- or camelcased).

So path of least resistance seemed to just stick to the cypress convention and accept that we might have some odd casing in env vars on CI. Benefit is that we'll have cypress config that aligns more closely with their docs. (Let me know if I've missed anything Viktor)

ismay:
> Base URL has a very specific meaning. It defines the URL to which end-point paths are appended to. This is the same reason that Cypress uses the term to identify the base URL of the site under test, so the consistency here is a good thing.

That sounds good. As long as the term means the same everywhere it shouldn't be confusing anyone 👍

varl:
0250f8e

For CI:

```
               uses: cypress-io/github-action@v2
               with:
                   start: yarn cypress:start
                   wait-on: 'http://localhost:3000'
                   wait-on-timeout: 300
               env:
                   CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
                   CYPRESS_dhis2BaseUrl: https://debug.dhis2.org/dev
                   CYPRESS_dhis2ApiVersion: 37
                   CYPRESS_networkMode: stub
```

Local:

```
cypress run --env networkMode=stub,dhis2BaseUrl=https://debug.dhis2.org/dev,dhis2ApiVersion=37
```

examples/testing-network-shim-app/cypress/support/index.js
varl: Done: d1adb66e255fe4330309c369d6d72ec8333b75e1

packages/cypress-commands/src/setups/enableAutoLogin.js
varl: d1adb66e255fe4330309c369d6d72ec8333b75e1

packages/cypress-plugins/src/plugins/networkShim/clearNetworkFixtures.js
varl: Agreed, we can refactor this later though as it is an internal detail. Leaving it for now.

varl:
@ismay & @Mohammer5 I have pushed changes to align this with the comments related to:

1. The Cypress enviroment vs. OS environment. This now adheres to the Cypress documentation, which we have concluded makes sense since `Cypress.env()` is only accessible in the browser, and `process.env` from the NodeJS Parts of Cypress. It is possible to bridge the gap in `plugins/index.js` and in the bootstrap code in `cypress-plugins`. The official approach works fine without additional code, so we use that. If we do need to deviate from that in the future, we have that option available.

2. Helpers for the network mode (`isStubMode`, `isCaptureMode`, `isLiveMode`) have been extracted and re-used instead of duplicating the checks in multiple places. Same for `getDhis2BaseUrl` and `setDhis2BaseUrlToLocalStorage`.

Mohammer5 (approved):

varl:
Rebased my commits to make sure we only get the "right" breaking changes: dc58462

PR #223, opened 2021-06-08
docs(network-shim): add guide explaining usage and configuration

HendrikThePendric:
I think these docs cover everything that's required. Writing docs is not something I do often, so please don't hesitate to correct me.

One thing I was wondering about is wether or not we need to add anything under `docs/setups`... The `enableNetworkShim` falls under `setups` too and is currently missing. However, I don't think it's very useful to document this in isolation. The function doesn't even accept any arguments so there is hardly anything to document.

On the other hand, it may be weird that we have two "setups" but only one of them has its own documentation page. Perhaps We just add a file that refers readers to the guide?

### DIFF
--- a/docs/guides/using-the-network-shim.md
+++ b/docs/guides/using-the-network-shim.md
@@ -1,0 +1,286 @@
+# Using the Network Shim
+
+#### Table of contents
+
+-   **[About](#about)**
+-   **[Quick setup](#quick-setup)**
+-   **[General workflow](#general-workflow)**
+-   **[Configuration](#configuration)**
+    -   **[Cypress](#cypress)**
+        -   **[Environment variables](#environment-variables)**
+        -   **[Plugins file](#plugins-file)**
+        -   **[Support file](#support-file)**
+    -   **[Package scripts and dependencies](#package-scripts-and-dependencies)**
+    -   **[GitHub workflow](#github-workflow)**
+
+## About
+
+The Network Shim makes it possible to record and play back network traffic during a cypress tests run. This means no live network requests need to be made when running e2e tests on CI, which is generally considered an anti-pattern as it can lead to flaky tests.
+
+It consists of a cypress plugin as well as a setup utility to be placed in cypress' support file. These have to be used in conjunction. Under the hood it uses the cypress `cy.intercept()` API to intercept network traffic during a test run. It has three different _modes_:
+
+-   `capture`: when operating in this mode, the Network Shim will allow all outgoing network traffic to the specified hosts and and store each response as a fixture (JSON file).
+-   `stub`: in this mode the Network Shim intercepts all outgoing requests to the specified hosts. It will reply with the responses from previously recorded fixtures.
+-   `live`: the Network Shim does not interfere with any network traffic when this mode is enabled. All network traffic is live.
+
+By recording fixtures during a _capture run_ and then using these fixtures during a _stub run_ a full cypress test suite can be executed without a live backend. This is especially useful when running cypress tests on CI. Since it is no longer necessary to have a network connection and a backend, the cypress tests on CI can be more stable. One caveat is that the tests are slightly less realistic, since the network traffic is mocked, so the developer will have to make sure that the recorded fixtures are kept up-to-date.
+
+A capture or stub run is executed for a specific DHIS2 api version and individual fixtures for each version are recorded/played-back during a run. This makes it easier to do cross-version testing for _feature-toggling_ apps.
+
+## Quick setup
+
+The command-line-tool in the package `@dhis2/cli-utils-cypress` comes with an `install` command which generates most of the configuration files documented below. See the [Setup Cypress for your project](https://cli-utils-cypress.dhis2.nu/#/guides/setting-up-cli-tool) guide for more info on how to do this.
+
+To get up-and-running quickly do the following:
+
+1.  Install the cli tool
+
+    ```bash
+    yarn add --dev @dhis2/cli-utils-cypress
+    ```
+
+    <br/>
+
+1.  Issue the install command to generate a working setup
+
+    ```bash
+    d2-utils-cypress install all
+    ```
+
+    (_Instead of `all` another "group" can be specified._)
+    <br/>
+
+1.  Ensure to git-ignore the `./cypress.env.json` file
+    <br/>
+
+1.  Install the peer dependency
+
+    ```bash
+    yarn add cypress
+    ```
+
+    (_Also install `cypress-cucumber` when applicable_)
+    <br/>
+
+1.  Install tools needed to run the scripts in `package.json`:
+    _(or install the `start-server-and-test` package instead)_
+
+    ```bash
+    yarn install -D wait-on concurrently
+    ```
+
+        <br/>
+
+1.  Create scripts `package.json` as demonstrated [here](#package-scripts-and-dependencies)
+    <br/>
+
+1.  Perform a capture run to generate fixtures and commit the generated files:
+
+    ```bash
+    yarn cy:capture
+    ```
+
+    <br/>
+
+1.  Copy-paste GitHub Actions job from [here](https://github.com/dhis2/workflows/blob/master/ci/dhis2-verify-app.yml#L73-L105) into the app's workflow to enable running e2e tests on CI
+    <br/>
+
+## General workflow
+
+Once the Network Shim is up-and-running, the general steps required get a cypress test-suite running on CI are as follows:
+
+1. Ensure a live backend is up and running
+1. Ensure a fully passing cypress test-suite is in place
+1. Ensure the CI workflow includes a job that runs the cypress test suite in stub mode
+1. Execute a capture run, which will produce fixture files
+1. Commit the generated fixture files and push to the remote
+1. The test suite will now execute on CI
+
+To setup cross-version testing essentially means taking some of the steps above multiple times:
+
+-   Perform multiple capture runs against different versions of the live backend, so network fixtures are created for all these versions
+-   Define multiple jobs in the GitHub workflow that execute a stub run against these different versions
+
+## Configuration
+
+A workflow as described above consists of a few moving parts, which are described in more details below. **However, running the `d2-utils-cypress install` command should create all required files**.
+
+### Cypress
+
+#### Environment variables
+
+The Network Shim depends on a few cypress environment variables. These can be set using the `--env` flag when interacting with the cypress CLI, but base values should be provided as well in via the following configuration files, see the [cypress docs](https://docs.cypress.io/guides/guides/environment-variables) for more information.
+
+##### `./cypress.json`
+
+This file is meant to be checked into source control, so should not contain any sensitive or machine-specific information.
+
+```json
+{
+    "baseUrl": "APP_URL",
+    "projectId": "PROJECT_ID",
+    "testFiles": "TEST_FILES",
+    "experimentalInteractiveRunEvents": true,
+    "video": false,
+    "env": {
+        "dhis2DataTestPrefix": "DATA_TEST_PREFIX",
+        "networkMode": "live", // default mode for the Network Shim
+        "dhis2ApiVersion": "value_from_prompt"
+    }
+}
+```
+
+_Notes:_
+
+-   _Only the `env` fields are required directly by the Network Shim, but other fields are included in this example since they are required to have a working cypress setup._
+-   _Interactive runs are not formally supported by the Network Shim, because `cypress open` (interactive mode) and `cypress run` have different behaviour regarding the plugin events. However, cypress is attempting to smooth things over and by setting `experimentalInteractiveRunEvents` to `true` it should theoretically be possible to at at least use stub mode in interactive mode._
+
+##### `./cypress.env.json`
+
+This file is meant to be git-ignored and we can use this to set local environment variable that may contain sensitive or machine-dependent information.
+
+```json
+{
+    "dhis2Username": "USERNAME", // secret
+    "dhis2Password": "PASSWORD", // secret
+    "dhis2BaseUrl": "DHIS2_BASE_URL" // machine specific
+}
+```
+
+_Note that `dhis2Username` and `dhis2Password` are not required directly by the Network Shim, but by the `enableAutologin` setup. They are included in this example since they will be needed for most use-cases._
+
+#### Plugins file
+
+The `networkShim` plugin should be imported and initialised in the cypress plugins file, see the [cypress docs](https://docs.cypress.io/guides/tooling/plugins-guide#Using-a-plugin) for more info. The plugin function doesn't require the second `config` argument most plugins accept.
+
+```js
+const { networkShim } = require('@dhis2/cypress-plugins')
+
+module.exports = (on, config) => {
+    networkShim(on)
+}
+```
+
+The `networkShim` plugin accepts a second `options` argument. For most DHIS2 platform apps the default options will be correct so there will be no need to pass a second argument.
+
+For most DHIS2 Cypress test suites the `enableAutoLogin` plugin can help reduce repetitive login steps in tests. In turn, the `enableAutoLogin` setup utility requires the `chromeAllowXSiteCookies` plugin when using Chrome. So a typical DHIS2 app will use this plugin alongside the `networkShim`.
+
+Below is an example of a plugins file with `options` supplied to the `networkShim` and used in conjunction with `chromeAllowXSiteCookies`:
+
+```js
+const {
+    networkShim,
+    chromeAllowXSiteCookies,
+} = require('@dhis2/cypress-plugins')
+
+const options = {
+    hosts: ['https://main-api.com', 'https://secondary-api.com'],
+    staticResources: [
+        '/endpoint-which-always-returns-the-same-data',
+        '/another-endpoint-which-always-returns-the-same-data',
+    ],
+}
+
+module.exports = (on, config) => {
+    networkShim(on, options)
+    chromeAllowXSiteCookies(on, config)
+}
+```
+
+#### Support file
+
+The `enableNetworkShim` setup should be imported and executed in the [cypress support file](https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests#Support-file). This function accepts no arguments. Typically, in a DHIS2 platform app, the `enableAutoLogin` setup is required so individual tests don't require login steps, which means the file should look like this:
+
+```js
+import { enableAutoLogin, enableNetworkShim } from '@dhis2/cypress-commands'
+
+enableAutoLogin()
+enableNetworkShim()
+```
+
+### Package scripts and dependencies
+
+##### `./package.json`
+
+The following `devDependencies` are required:
+
+```json
+"devDependencies": {
+    "@dhis2/cypress-commands": "^#.#.#",
+    "@dhis2/cypress-plugins": "^#.#.#",
+},
+```
+
+And it is recommended to add some convenience commands to the `scripts` section, for example:
+
+```json
+"scripts": {
+    "cy:start": "BROWSER=none yarn start",
+    "cy:open": "wait-on 'http-get://localhost:3000' && cypress open",
+    "cy:run": "wait-on 'http-get://localhost:3000' && cypress run",
+    "cy:capture": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=capture' --kill-others --success first",
+    "cy:stub": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=stub' --kill-others --success first",
+},
+```
+
+_Note that `cy:stub` here is meant to perform a local stub-run, the stub run on CI should be configured differently as explained in the [GitHub workflow](#github-workflow) section._
+
+Some notes:
+
+1. These example scripts assume both `wait-on` and `concurrently` are installed available. If this is not the case, run:
+    ```bash
+    yarn add -D wait-on concurrently
+    ```
+1. As an alternative to using `wait-on` and `concurrently` the package `start-server-and-test` could be used, which is recommended by cypress. An example can be found [here](https://github.com/dhis2/scheduler-app/blob/8af8ae7b8b319384f14b510b7017f2beea730f5c/package.json#L14-L16).
+1. Setting `BROWSER=none` before running `yarn start` simply means that no browser tab is going to be opened/activated once the app is ready.
+1. In the example above no `dhis2ApiVersion` is passed as a cypress env variable, even though the Network Shim relies on this. This works because that variable is declared in `./cypress.env.json` so that value is used. See the [cypress docs](https://docs.cypress.io/guides/guides/environment-variables) for more info.
+1. The example above only takes into account testing against a single backend version. If an additional v36 version of DHIS2 core was available on `http://localhost:8081` and the test-suite should be executed against that backend as well, the following scripts could be added:
+    ```json
+    "cy:capture-v36": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=capture,dhis2BaseUrl=http://localhost:8081,dhis2ApiVersion=36' --kill-others --success first",
+    "cy:stub-v36": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=stub,dhis2BaseUrl=http://localhost:8081,dhis2ApiVersion=36' --kill-others --success first",
+    ```
+
+### GitHub workflow
+
+At DHIS2 we use GitHub Actions for CI/CD, and we provide a workflow template for DHIS2 platform apps [here](https://github.com/dhis2/workflows/blob/master/ci/dhis2-verify-app.yml#L73-L105). A typical DHIS2 platform-app will probably use the full `dhis2-verify-app` workflow and store this in `./.github/workflows/dhis2-verify-app.yml`. Once the Network Shim has been configured and a capture run has been executed, all that is needed is to uncomment the e2e job.
+
+Below is an illustration of such a job to provide some additional info:
+
+-   The job uses the `cypress-io/github-action@v2` GitHub action.
+-   The required environment variables are set via the `env` parameter, note that:
+    -   They need to be prefixed with `CYPRESS_`.
+    -   The `dhis2Username` and `dhis2Password` variables are not needed because we only do a stub run on CI.
+-   To perform cross-version testing, more steps like this could be added, specifying a different value for `CYPRESS_dhis2BaseUrl` and `CYPRESS_dhis2ApiVersion`.
+-   In the example below `record` and `parallel` have been enabled. This will only work if the project has been registered at the cypress dashboard and has a valid record key.
+
+```yml
+e2e:
+    runs-on: ubuntu-latest
+    if: "!github.event.push.repository.fork && github.actor != 'dependabot[bot]'"
+    strategy:
+        fail-fast: false
+        matrix:
+            containers: [1, 2, 3, 4]
+    steps:
+        - name: Checkout
+            uses: actions/checkout@v2
+        - uses: actions/setup-node@v1
+            with:
+                node-version: 12.x
+        - name: End-to-End tests
+            uses: cypress-io/github-action@v2
+            with:
+                start: yarn cy:start
+                wait-on: 'http://localhost:3000'
+                wait-on-timeout: 300
+                record: true
+                parallel: true
+            env:
+                GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+                CYPRESS_dhis2BaseUrl: http://localhost:8080
+                CYPRESS_dhis2ApiVersion: 37
+                CYPRESS_networkMode: stub
+```
+
+Github actions were used as an example in these docs, but it is also possible to setup cypress test-runs with these plugins on other CI solutions.
--- a/docs/_sidebar.md
+++ b/docs/_sidebar.md
@@ -4,6 +4,7 @@
     -   [**Enabling auto login**](guides/enable-auto-login)
     -   [**Custom data test syntax**](guides/custom-data-test-syntax)
     -   [**Adding login credentials**](guides/add-login-credentials)
+    -   [**Using the network shim**](guides/using-the-network-shim)
 -   [**API: Commands**](commands)
     -   [**getWithDataTest**](commands/getWithDataTest)
     -   [**findWithDataTest**](commands/findWithDataTest)
--- a/examples/platform-app/package.json
+++ b/examples/platform-app/package.json
@@ -13,8 +13,8 @@
         "format": "d2-style apply",
         "format:staged": "yarn format --staged",
         "cy:start": "BROWSER=none yarn start",
-        "cy:open": "wait-on 'http://localhost:3000' && cypress open",
-        "cy:run": "wait-on 'http://localhost:3000' && cypress run",
+        "cy:open": "wait-on 'http-get://localhost:3000' && cypress open",
+        "cy:run": "wait-on 'http-get://localhost:3000' && cypress run",
         "cy:capture": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=capture' --kill-others --success first",
         "cy:stub": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=stub' --kill-others --success first",
         "cy:capture-and-stub": "yarn cy:capture && yarn cy:stub"
--- a/examples/testing-network-shim-app/package.json
+++ b/examples/testing-network-shim-app/package.json
@@ -21,8 +21,8 @@
         "test": "react-scripts test",
         "eject": "react-scripts eject",
         "cy:start": "BROWSER=none yarn start",
-        "cy:open": "wait-on 'http://localhost:3000' && cypress open",
-        "cy:run": "wait-on 'http://localhost:3000' && cypress run",
+        "cy:open": "wait-on 'http-get://localhost:3000' && cypress open",
+        "cy:run": "wait-on 'http-get://localhost:3000' && cypress run",
         "cy:capture": "concurrently 'yarn serve' 'yarn cy:start' 'yarn cy:run --env networkMode=capture' --kill-others --success first",
         "cy:stub": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=stub' --kill-others --success first",
         "cy:capture-and-stub": "yarn cy:capture && yarn cy:stub"
--- a/yarn.lock
+++ b/yarn.lock
@@ -4075,7 +4075,7 @@ ansi-escapes@^4.2.1:
   dependencies:
     type-fest "^0.11.0"
 
-ansi-escapes@^4.3.0, ansi-escapes@^4.3.1:
+ansi-escapes@^4.3.1:
   version "4.3.2"
   resolved "https://registry.yarnpkg.com/ansi-escapes/-/ansi-escapes-4.3.2.tgz#6b2291d1db7d98b6521d5f1efa42d0f3a9feb65e"
   integrity sha512-gKXj5ALrKWQLsYG9jlTRmR/xKluxHV+Z9QEwNIgCfM1/uwPMCuzVVnh5mwTd+OuBZcwSIMbqssNWRm1lE51QaQ==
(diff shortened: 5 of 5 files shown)


docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+    "projectId": "PROJECT ID",
+    "testFiles": "TEST FILES",
+    "experimentalInteractiveRunEvents": true,
+    "video": false
+}
+```
+
+#### `./cypress/plugins/index.js`
+
+In order to execute a cypress run in a DHIS2 platform app, the `enableAutoLogin` setup is required so the user is authorised to access the app. In turn, the `enableAutoLogin` setup for users on Chrome requires the `chromeAllowXSiteCookies`. As such, it makes sense to always use this plugin alongside the `networkShim`.
```
Mohammer5, line 69:
```suggestion
For most DHIS2 cypress setups, using the `enableAutoLogin` setup helps managing authentication conveniently. In turn, the `enableAutoLogin` setup utility requires the `chromeAllowXSiteCookies` when using Chrome. In most cases it makes sense to always use this plugin alongside the `networkShim`.
```

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+ * }
+ */
+
+module.exports = (on, config) => {
+    networkShim(on /*, networkShimOptions */)
+    chromeAllowXSiteCookies(on, config)
+}
+```
+
+#### `./cypress/support/index.js`
```
Mohammer5, line 98:
```suggestion
#### Cypress' `supportFile`
```

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+
+The `enableNetworkShim` setup accepts no arguments and in order to execute a cypress run in a DHIS2 platform app, the `enableAutoLogin` setup is required, which means the file should look like this:
```
Mohammer5, line 100:
```suggestion
The `enableNetworkShim` setup accepts no arguments and in order to execute a cypress run in a DHIS2 platform app. It's recommended that most apps use the `enableAutoLogin` setup utility to manage automatic authentication, which means the file should look like this:
```

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Package scripts and dependencies
+
+#### `./package.json`
+
+The following `devDependencies` are required:
+
+```json
+"devDependencies": {
+    "@dhis2/cli-utils-cypress": "^#.#.#",
```
Mohammer5, line 117: This one is not required anymore, as `cypress` and `cypress-cucumber-preprocessor` are now peer dependencies and are being used directly

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+},
+```
+
+And it is recommended to add some convenience commands to the `scripts` section, for example:
+
+```json
+"scripts": {
+    "cy:start": "BROWSER=none yarn start",
+    "cy:open": "wait-on 'http://localhost:3000' && cypress open",
+    "cy:run": "wait-on 'http://localhost:3000' && cypress run",
```
Mohammer5, line 129:
```suggestion
    "cy:open": "wait-on 'http-get://localhost:3000' && cypress open",
    "cy:run": "wait-on 'http-get://localhost:3000' && cypress run",
```

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+    "cy:capture": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=capture' --kill-others --success first",
+    "cy:stub": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=stub' --kill-others --success first",
```
Mohammer5, line 131:
```suggestion
    "cy:capture": "concurrently 'yarn cy:run --env networkMode=capture' 'yarn cy:start' --kill-others --success first",
    "cy:stub": "concurrently 'yarn cy:run --env networkMode=stub' 'yarn cy:start' --kill-others --success first",
```

`--success first` means: The command executes successfully when the first command passed to `concurrently` exits with code 0

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+    yarn add -D wait-on concurrently
+    ```
+1. Setting `BROWSER=none` before running `yarn start` simply means that no browser tab is going to be opened/activated once the app is ready.
+1. In the example above no `dhis2ApiVersion` is passed as a cypress env variable, even though the Network Shim relies on this. This works because that variable is declared in `./cypress.env.json` so that value is used. See the [cypress docs](https://docs.cypress.io/guides/guides/environment-variables) for more info.
+1. The example above only takes into account testing against a single backend version. Suppose a v37 version of DHIS2 core was available on `http://localhost:8080` and a v36 version on `http://localhost:8081`, the package scripts could look like this:
+    ```json
+    "scripts": {
+        "cy:start": "BROWSER=none yarn start",
+        "cy:open": "wait-on 'http://localhost:3000' && cypress open",
+        "cy:run": "wait-on 'http://localhost:3000' && cypress run",
```
Mohammer5, line 148:
```suggestion
        "cy:open": "wait-on 'http-get://localhost:3000' && cypress open",
        "cy:run": "wait-on 'http-get://localhost:3000' && cypress run",
```

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+        "cy:capture-v37": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=capture' --kill-others --success first",
+        "cy:stub-v37": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=stub' --kill-others --success first",
+        "cy:capture-v36": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=capture,dhis2BaseUrl=http://localhost:8081,dhis2ApiVersion=36' --kill-others --success first",
+        "cy:stub-v36": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=stub,dhis2BaseUrl=http://localhost:8081,dhis2ApiVersion=36' --kill-others --success first",
```
Mohammer5, line 152:
```suggestion
        "cy:capture-v36": "concurrently 'yarn cy:run --env networkMode=capture,dhis2BaseUrl=http://localhost:8081,dhis2ApiVersion=36' 'yarn cy:start' --kill-others --success first",
        "cy:stub-v36": "concurrently 'yarn cy:run --env networkMode=stub,dhis2BaseUrl=http://localhost:8081,dhis2ApiVersion=36' 'yarn cy:start' --kill-others --success first",
```

See comment above regarding these commands with `--success first`

Mohammer5 (changes requested):

HendrikThePendric:
Once https://jira.dhis2.org/browse/CLI-54 is done, these docs will need to be updated so `./cypress.json` and `./cypress.env.json` match the generated files.

docs/guides/using-the-network-shim.md
Mohammer5: I was wrong: `--success first` means that concurrently will exit with 0 or 1 based on the result of the first command that finishes, regardless of the order

docs/guides/using-the-network-shim.md
Mohammer5: I was wrong: `--success first` means that concurrently will exit with 0 or 1 based on the result of the first command that finishes, regardless of the order

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+
+#### `./cypress.env.json`
+
+The network shim expects a few environment variables to be present. These can be set using the `--env` flag when interacting with the cypress CLI, but base values should be provided as well the `./cypress.env.json` configuration file:
+
+```json
+{
+    "dhis2Username": "YOUR_USERNAME_HERE",
+    "dhis2Password": "YOUR PASSWORD HERE",
+    "dhis2BaseUrl": "http://localhost:8080",
```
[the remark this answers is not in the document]
HendrikThePendric: `dhis2BaseUrl` is required by the NetworkShim too, but you are right about the other properties. Personally I think it's OK to keep them in, since I already mention that in order to use the Network Shim you will also need the `enableAutoLogin` setup and the `chromeAllowXSiteCookies` plugin.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
+# Using the Network Shim
+
+## About
+
+The Network Shim is a tool that provides a cypress plugin as well as a setup utilility for cypress' `setupFile`, which have to be used in conjunction. It uses the cypress `cy.intercept()` API to intercept network traffic during a test run. It has three different _modes_:
```
ismay, line 5: Maybe you could start with a succinct explanation of what the network shim is meant to be used for. Like:

> The network shim allows you to record and play back network traffic during your tests. This means you don't have to make live network requests when running e2e tests on CI, which is generally considered an anti-pattern as it can lead to flaky tests.

docs/guides/using-the-network-shim.md
HendrikThePendric: resolved via 00f193e

docs/guides/using-the-network-shim.md
HendrikThePendric: resolved via 00f193e

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+
+-   `capture`: when operating in this mode, the Network Shim will intercept all outgoing requests to the specified hosts and allow each request to reach the destination server. Each response is stored as a fixture (JSON file).
+-   `stub`: in this mode the Network Shim also intercepts all outgoing requests to the specified hosts, but it will prevent the request from reaching the destination server and instead it will produce a response from a previously recorded fixture.
```
ismay, line 8: Instead of produce:

> it will produce a response from a previously recorded fixture.

This seems easier to understand to me:

> it will reply with a response from a previously recorded fixture.

Also the sentence for this bullet point is a bit long. I'd split it up into multiple for readability.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+-   `live`: the Network Shim does not intercept any requests when this mode is enabled.
+
+By first recording fixtures during a _capture run_ and then using these fixtures during a _stub run_ it effectively becomes possible to execute a full cypress test suite without a live backend. This is especially relevant for running a cypress test suite on CI.
```
ismay, line 11: This relates to the intro text I suggested. Couple of tweaks:

> By recording fixtures during a _capture run_ and then using these fixtures during a _stub run_ you can execute a full cypress test suite without a live backend. This is especially useful when running cypress tests on CI. Since you no longer have to rely on a network connection and a backend, your cypress tests on CI can be more stable. One caveat is that your tests are slightly less realistic, since the network traffic is mocked, so you'll have to make sure that your recorded fixtures are kept up-to-date.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+
+A capture or stub run is executed for a specific DHIS2 api version and individual fixtures for each version are produced/used during a run. This paves the way for doing cross-version testing for _feature-toggling_ apps.
```
ismay, line 13: `produced/used` -> `recorded/played back`

The record/playback analogy seems easier to grasp to me

ismay: Maybe:

`This paves the way for doing cross-version ...` -> `This makes it easier to do cross-version ...`

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+
+## General workflow
+
+As mentioned, he Network Shim is a tool that enables a workflow where a cypress e2e test suite is executed on CI.
```
ismay, line 17: This is a little vague maybe. I'd just mention that it's a tool that allows you to record, play back network traffic. And potentially also that that allows for more stable e2e tests on CI. 

Because technically without it you can still run e2e tests on CI. It's just that they'll be more flaky.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+
+The general steps required to setup a workflow like that are as follows:
```
ismay, line 19: Maybe for the workflow part of these steps we can refer to our workflows repo.

ismay: Also I guess people don't necessarily have to use GH actions. We could mention that that's what we use, but you can use any CI you'd like.

docs/guides/using-the-network-shim.md
ismay: I think I'd put the succinct explanation completely at the top. Then followed by an install command (if that's necessary). And that followed by a very short TLDR that lists the simplest, shortest way to get started. And then after that we can go more in depth.

That to me often is the most readable manner to document. Basically going from simple/high level -> complex/detailed.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Configuration
+
+A workflow as described above consists of a few moving parts, which are described in more details below. **However, running the `d2-utils-cypress install` command should create all required files**.
+
+### Cypress
+
+#### `./cypress.env.json`
+
+The network shim expects a few environment variables to be present. These can be set using the `--env` flag when interacting with the cypress CLI, but base values should be provided as well the `./cypress.env.json` configuration file:
```
ismay, line 41: Maybe we can just link to the cypress environment variable docs to show there are several ways to provide environment variables, and then just list a single method here (the one we prefer/recommend).

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+    "dhis2Password": "YOUR PASSWORD HERE",
+    "dhis2BaseUrl": "http://localhost:8080",
+    "networkMode": "live",
+    "dhis2ApiVersion": "37"
+}
+```
+
+#### `./cypress.json`
+
+Interactive runs are not formally supported by the Network Shim, because `cypress open` (interactive mode) and `cypress run` have different behaviour regarding the plugin events. However, cypress is attempting to smooth things over and by setting `experimentalInteractiveRunEvents` to true it should theoretically be possible to at at least use stub mode in interactive mode, for example:
```
ismay, line 55: This is a little unclear to me. If I understand what you're saying correctly, maybe you can phrase it as:

> Stubbing partial runs is not supported

I think I wouldn't even go in depth in documenting something that we don't officially support. Just mentioning that we don't support it seems good enough to me. Otherwise maybe it'll only get confusing.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+ *         '/another-endpoint-which-always-return-the-same-data',
+ *     ]
+ * }
+ */
+
+module.exports = (on, config) => {
+    networkShim(on /*, networkShimOptions */)
+    chromeAllowXSiteCookies(on, config)
+}
+```
```
ismay, line 96: This code block could use a little more context for me. It doesn't immediately follow from the paragraph above what the user is supposed to do with this code block. Also the commented out code seems confusing as it seems your demonstrating two usecases at once. I'd separate the simple one and the elaborate on, starting with the simple one that should apply in most cases.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+
+#### `./cypress/support/index.js`
+
+The `enableNetworkShim` setup accepts no arguments and in order to execute a cypress run in a DHIS2 platform app. It's recommended that most apps use the `enableAutoLogin` setup utility to manage automatic authentication, which means the file should look like this:
```
ismay, line 100: `The `enableNetworkShim` setup accepts no arguments and in order to execute a cypress run in a DHIS2 platform app.`

I think this sentence got a little mangled. Should it be something like:

> The `enableNetworkShim` setup does not require any arguments in order to execute a cypress run in a DHIS2 platform app.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+```js
+import { enableAutoLogin, enableNetworkShim } from '@dhis2/cypress-commands'
+
+enableAutoLogin()
+enableNetworkShim()
+```
+
+### Package scripts and dependencies
+
+#### `./package.json`
```
ismay, line 111: I'd move this to the top, to an `install` section (as I mentioned in a comment above)

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+
+And it is recommended to add some convenience commands to the `scripts` section, for example:
+
+```json
+"scripts": {
+    "cy:start": "BROWSER=none yarn start",
+    "cy:open": "wait-on 'http-get://localhost:3000' && cypress open",
+    "cy:run": "wait-on 'http-get://localhost:3000' && cypress run",
+    "cy:capture": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=capture' --kill-others --success first",
+    "cy:stub": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=stub' --kill-others --success first",
```
ismay, line 131: Personally I'd go with the start-server-and-test module. Cypress recommends it in their docs, it's maintained by a cypress maintainer, and it seems easier to grasp to a newcomer to me.

Concurrently requires more in-depth knowledge to use correctly.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,193 @@
...
+1. The example above only takes into account testing against a single backend version. Suppose a v37 version of DHIS2 core was available on `http://localhost:8080` and a v36 version on `http://localhost:8081`, the package scripts could look like this:
+    ```json
+    "scripts": {
+        "cy:start": "BROWSER=none yarn start",
+        "cy:open": "wait-on 'http-get://localhost:3000' && cypress open",
+        "cy:run": "wait-on 'http-get://localhost:3000' && cypress run",
+        "cy:capture-v37": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=capture' --kill-others --success first",
+        "cy:stub-v37": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=stub' --kill-others --success first",
+        "cy:capture-v36": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=capture,dhis2BaseUrl=http://localhost:8081,dhis2ApiVersion=36' --kill-others --success first",
+        "cy:stub-v36": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=stub,dhis2BaseUrl=http://localhost:8081,dhis2ApiVersion=36' --kill-others --success first",
```
ismay, line 152: I think the scripts code sections are getting a little bulky. Maybe we could cut down on the amount of code here, say by linking to a codesandbox, an example app in this repo, or something like that.

ismay:
I commented on the specifics that stood out to me, but as a general I note I think I'd go with the following structure: 

1. Start with the happy path, keep that as simple and succinct as possible
2. Then branch out into specifics, exceptions, etc.

Sometimes I've noticed the docs immediately go into the possible exceptions/config/etc., and that makes it a lot harder to read.

docs/guides/using-the-network-shim.md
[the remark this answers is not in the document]
HendrikThePendric: Resolving this, as it was discussed on slack

docs/guides/using-the-network-shim.md
HendrikThePendric: Yeah... I struggled with this a little bit... I like having the `experimentalInteractiveRunEvents ` toggled to `true` in the config because it makes it more likely that things will work correctly when using `cypress open` instead of `cypress run`. So I'd like to keep that field in.... But since it's there, staring people in the face, I felt like I should also provide an explanation. And that's how I ended up writing this. Would you recommend just removing the explanation?

docs/guides/using-the-network-shim.md
HendrikThePendric: Not sure what happened to that sentence, thanks for the suggestion, but all I tried to say was that you don't need to supply any arguments to the function.

docs/guides/using-the-network-shim.md
HendrikThePendric: I'll keep it here because I have something else in mind for that install section you mention.

docs/guides/using-the-network-shim.md
HendrikThePendric: I'll add it as a recommendation below

docs/guides/using-the-network-shim.md
HendrikThePendric: Tweaked that a bit so we only have two lines now.

docs/guides/using-the-network-shim.md
HendrikThePendric: I think this is a good suggestion but I don't think I can implement it exactly....
1. Is sort of done already by tweaking the about section
2. Doesn't completely exists, but the `cli-utils-cypress` CLI does indeed come with an install command, which should generate (most of) the required files.
3. I see 2 & 3 being overlapping in our case, I'll probably end up combining these
4. I think I've got the in-depth part covered 🙄 
5. I will just create some clickable list leading to the headers

I'll ping you when I'm done

Mohammer5 (approved):

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+-   `stub`: in this mode the Network Shim also intercepts all outgoing requests to the specified hosts. However, it will prevent the request from reaching the destination server and instead it will reply with a response from a previously recorded fixture.
+-   `live`: the Network Shim does not intercept any requests when this mode is enabled.
+
+By first recording fixtures during a _capture run_ and then using these fixtures during a _stub run_ it effectively becomes possible to execute a full cypress test suite without a live backend. This is especially relevant for running a cypress test suite on CI.
+
+By recording fixtures during a _capture run_ and then using these fixtures during a _stub run_ a full cypress test suite can be executed without a live backend. This is especially useful when running cypress tests on CI. Since it is no longer necessary to have a network connection and a backend, the cypress tests on CI can be more stable. One caveat is that the tests are slightly less realistic, since the network traffic is mocked, so the developer will have to make sure that the recorded fixtures are kept up-to-date.
```
ismay, line 28: These two paragraphs seem quite similar.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+    -   **[Package scripts and dependencies](#package-scripts-and-dependencies)**
+    -   **[GitHub workflow](#github-workflow)**
+
+## About
+
+The Network Shim makes it possible to record and play back network traffic during a cypress tests run. This means no live network requests need to be made when running e2e tests on CI, which is generally considered an anti-pattern as it can lead to flaky tests.
+
+It consists of a cypress plugin as well as a setup utility to be placed in cypress' support file. These have to be used in conjunction. Under the hood it uses the cypress `cy.intercept()` API to intercept network traffic during a test run. It has three different _modes_:
+
+-   `capture`: when operating in this mode, the Network Shim will intercept all outgoing requests to the specified hosts and allow each request to reach the destination server. Each response is stored as a fixture (JSON file).
```
ismay, line 22: Maybe we could clarify this a little by wording it like so:

> when operating in this mode, the Network Shim will allow all outgoing network traffic to the specified hosts and and store each response as a fixture (JSON file).

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+-   `stub`: in this mode the Network Shim also intercepts all outgoing requests to the specified hosts. However, it will prevent the request from reaching the destination server and instead it will reply with a response from a previously recorded fixture.
```
ismay, line 23: I found the "also" a little confusing here:

> in this mode the Network Shim intercepts all outgoing requests to the specified hosts. It will reply with the responses from previously recorded fixtures.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+-   `live`: the Network Shim does not intercept any requests when this mode is enabled.
```
ismay, line 24: I was thinking this might clarify it a little:

> the Network Shim does not interfere with any network traffic when this mode is enabled. All network traffic is live.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+
+    <br/>
+
+1.  Issue the install command to generate a working setup
+
+    ```bash
+    d2-utils-cypress install all
+    ```
+
+    (_Instead of `all` another "group" can be specified._)
```
ismay, line 52: Maybe link to where users can find out what those groups are, and what they (the groups) mean.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+    ```
+
+    <br/>
+
+1.  Copy-paste GitHub Actions job from [here](https://github.com/dhis2/workflows/blob/master/ci/dhis2-verify-app.yml#L73-L105) into the app's workflow to enable running e2e tests on CI
+    <br/>
+
+## General workflow
+
+Once Network Shim is up-and-running, the general steps required get a cypress test-suite running on CI are as follows:
```
ismay, line 92: Once `the` Network Shim

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+
+1. Ensure a live backend is up and running
+1. Ensure a fully passing cypress test-suite is in place
+1. Ensure the CI workflow includes a job that runs the cypress test suite in stub mode
+1. Execute a capture run, which will produce fixture files
+1. Commit the generated fixture files and push to the remote
+1. The test suite will now execute on CI and if a change breaks a test it will be impossible to merge
```
ismay, line 99: I think I'd remove this: `and if a change breaks a test it will be impossible to merge`. Seems a little confusing to me, and the end goal is to get the suite to run on CI, so I think it's fine to stop there. Plus they might not block merging for failed tests, depends on their setup.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+```
+
+_Notes:_
+
+-   _Only the `env` fields are required directly by the Network Shim, but other fields are included in this example since they are required to have a working cypress setup._
+-   _Interactive runs are not formally supported by the Network Shim, because `cypress open` (interactive mode) and `cypress run` have different behaviour regarding the plugin events. However, cypress is attempting to smooth things over and by setting `experimentalInteractiveRunEvents` to `true` it should theoretically be possible to at at least use stub mode in interactive mode._
+
+##### `./cypress.env.json`
+
+This file is meant to be git-ignored and we can use this to set environment variable that may contain sensitive or machine-dependent information.
```
ismay, line 142: I'd specify: set `local` environment variable`s`

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+    "dhis2Password": "PASSWORD", // secret
+    "dhis2BaseUrl": "DHIS2_BASE_URL" // machine specific
+}
+```
+
+_Note that `dhis2Username` and `dhis2Password` are not required directly by the Network Shim, but by the `enableAutologin` setup. They are included in this example since they will be needed for most use-cases._
+
+#### Plugins file
+
+The `networShim` plugin should be imported and initialised in the cypress plugins file, see the [cypress docs](https://docs.cypress.io/guides/tooling/plugins-guide#Using-a-plugin) for more info. The plugin function doesn't require the second `config` argument most plugins accept.
```
ismay, line 156: networShim -> networkShim

docs/guides/using-the-network-shim.md
HendrikThePendric: I don't think that's needed, that link is already in the section intro.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+
+```js
+const { networkShim } = require('@dhis2/cypress-plugins')
+
+module.exports = (on, config) => {
+    networkShim(on)
+}
+```
+
+The `networkShim` plugin also accepts a second `options` argument. For most DHIS2 platform apps the default options will be correct so there will be no need to pass a second argument.
```
ismay, line 166: I'd remove `also`

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+
+Also for most DHIS2 cypress setups, using the `enableAutoLogin` setup helps managing authentication conveniently. In turn, the `enableAutoLogin` setup utility requires the `chromeAllowXSiteCookies` when using Chrome. So in most cases it makes sense to always use this plugin alongside the `networkShim`.
```
ismay, line 168: Tweaked it a little:

> For most DHIS2 Cypress test suites the `enableAutoLogin` plugin can help reduce repetitive login steps in tests. The `enableAutoLogin` setup utility requires the `chromeAllowXSiteCookies` flag to be set when using Chrome for testing.

This sentence:

> So in most cases it makes sense to always use this plugin alongside the `networkShim`.

Doesn't follow clearly for me from the sentences before it. Why does that make sense (to always use this plugin)?

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+const {
+    networkShim,
+    chromeAllowXSiteCookies,
+} = require('@dhis2/cypress-plugins')
+
+const options = {
+    hosts: ['https://main-api.com', 'https://secondary-api.com'],
+    staticResources: [
+        '/endpoint-which-always-return-the-same-data',
+        '/another-endpoint-which-always-return-the-same-data',
```
ismay, line 182: return`s`

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+
+module.exports = (on, config) => {
+    networkShim(on, options)
+    chromeAllowXSiteCookies(on, config)
+}
+```
+
+#### Support file
+
+The `enableNetworkShim` setup should be imported and executed in the [cypress support file](https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests#Support-file). This function accepts no arguments so should simply be called like this `enableNetworkShim()`. Typically, in a DHIS2 platform app, the `enableAutoLogin` setup is also required so individual tests don't require login steps, which means the file should look like this:
```
ismay, line 194: I'd do `like this` -> `with`

ismay: And I'd remove `also` from `setup is also`

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+
+_Note that `cy:stub` here is meant to perform a local stub-run, the stub run on CI should be configured differently as explained in the [GitHub workflow](#github-workflow) section._
+
+Some notes:
+
+1. These example scripts assume both `wait-on` and `concurrently` are installed available. If this is not the case, run:
+    ```bash
+    yarn add -D wait-on concurrently
+    ```
+1. As an alternative to using `wait-on` and `concurrently` the package `start-server-and-test` could be used, which is recommended by cypress. An example can be found [here](https://github.com/dhis2/scheduler-app/blob/master/package.json#L14-L16).
```
ismay, line 236: Maybe we should link to a specific point in time, so it's stable: https://github.com/dhis2/scheduler-app/blob/8af8ae7b8b319384f14b510b7017f2beea730f5c/package.json#L14-L16. Also, feel free to omit this if you think it's confusing and everyone should use wait-on and concurrently.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+1. The example above only takes into account testing against a single backend version. If an additional v36 version of DHIS2 core was available on `http://localhost:8081` and the test-suite should be executed against that backend as well, the following scripts could be added:
+    ```json
+    "cy:capture-v36": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=capture,dhis2BaseUrl=http://localhost:8081,dhis2ApiVersion=36' --kill-others --success first",
+    "cy:stub-v36": "concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=stub,dhis2BaseUrl=http://localhost:8081,dhis2ApiVersion=36' --kill-others --success first",
+    ```
+
+### GitHub workflow
+
+At DHIS2 we use GitHub Actions for CI/CD, and we provide a workflow template for DHIS2 platform apps [here](https://github.com/dhis2/workflows/blob/master/ci/dhis2-verify-app.yml#L73-L105). A typical DHIS2 platform-app will probably use the full `dhis2-verify-app` workflow and store this in `./.github/workflows/dhis2-verify-app.yml`. Once the Network Shim has been configured and a capture run has been exectued to create the network fixtures all that is needed is to uncomment the e2e job.
```
ismay, line 247: I'm thinking maybe we should link to a specific commit here as well. To ensure it's stable. And maybe then a link to the repo, stating: `Our workflows are subject to change. For the latest version of our workflows see: ...`

ismay: exectued -> executed

ismay: Was little hard to process because it's quite a long sentence, condensed it a little:

> Once the Network Shim has been configured and a capture run has been executed, all that is needed is to uncomment the e2e job.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+
+Below is an illustration of such a job to provide some additional info:
+
+-   The job is using the `cypress-io/github-action@v2` GitHub action.
```
ismay, line 251: is using -> uses

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+-   The required environment variables are being set via the `env` parameter, note that:
```
ismay, line 252: ~~being~~

docs/guides/using-the-network-shim.md
HendrikThePendric: Good catch

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
...
+                parallel: true
+            env:
+                GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+                CYPRESS_dhis2BaseUrl: http://localhost:8080
+                CYPRESS_dhis2ApiVersion: 37
+                CYPRESS_networkMode: stub
+```
+
+It should also be possible to setup cypress test-runs on other CI solutions, but no documentation about this is available.
```
ismay, line 288: Reworded it a little:

> Github actions were used as an example in these docs, but it is also possible to setup cypress test-runs with these plugins on other CI solutions.

docs/guides/using-the-network-shim.md
```diff
@@ -0,0 +1,288 @@
+# Using the Network Shim
+
+#### Table of content
```
ismay, line 3: content`s`

docs/guides/using-the-network-shim.md
ismay: I think the changes you've made have resolved this for me

docs/guides/using-the-network-shim.md
ismay: Ok, yeah I find that nice personally. Just a `see here` link makes it very clear where you can find the info when skimming.

docs/guides/using-the-network-shim.md
HendrikThePendric: Used a bit of mine and a bit of your suggestion here in the end.

docs/guides/using-the-network-shim.md
HendrikThePendric: Good suggestion, reg. the link, better to link to a snapshot.

Reg. `start-server-and-test`: I definitely don't think I'd want to omit that given the fact it's simpler and recommended by cypress. Instead we should probably switch the example apps to using that and subsequently update the docs to *only* mention that. Since we'd have to also update the example apps, it would bloat the scope of this PR a bit though and that's why I didn't want to do it now.

I've created follow-up issue [CLI-57](https://jira.dhis2.org/browse/CLI-57) in Jira.

PR #234, opened 2021-06-10
fix(install cmd): use correct env vars for network shim in install command

Mohammer5:


HendrikThePendric (approved):
I approve of your changes, and mine too 😅

PR #262, opened 2021-08-23
fix(network-shim): ensure 304s are dealth with correctly

HendrikThePendric:
This is a follow-up on #261. I guess the approval app test suite is the first place where we had to deal with repeating requests to the same endpoint that would return a mixture of 304 and 200 responses. In any case, I uncovered 2 bugs. I'll place some comments in the code to explain.

Before merging this in, I think I'd like to check what would happen if we apply these changes in isolation, without the changes from #261. I _think_ both PRs are required, but it could also be possible that the changes in the current PR are actually all that was needed.

### DIFF
--- a/packages/cypress-commands/src/setups/enableNetworkShim/captureRequests.js
+++ b/packages/cypress-commands/src/setups/enableNetworkShim/captureRequests.js
@@ -166,8 +166,9 @@ function processDuplicatedRequest({
         // RequestStub was already nonDeterministic, responseBody is already an array
         const matchingResponseBodyIndex = isNotModified
             ? // When isNotModified the response body is empty and the browser is expected
-              // to return the last known response, so we do the same.
-              requestStub.responseBody.lenght - 1
+              // to return the last known response, so we do the same by returning the 
+              // value of the last responseLookup.
+              requestStub.responseLookup[requestStub.responseLookup.length - 1]
             : requestStub.responseBody.findIndex(
                   responseBody => responseBody === newResponseBody
               )
--- a/packages/cypress-commands/src/setups/enableNetworkShim/captureRequests.js
+++ b/packages/cypress-commands/src/setups/enableNetworkShim/captureRequests.js
@@ -166,7 +166,7 @@ function processDuplicatedRequest({
         // RequestStub was already nonDeterministic, responseBody is already an array
         const matchingResponseBodyIndex = isNotModified
             ? // When isNotModified the response body is empty and the browser is expected
-              // to return the last known response, so we do the same by returning the 
+              // to return the last known response, so we do the same by returning the
               // value of the last responseLookup.
               requestStub.responseLookup[requestStub.responseLookup.length - 1]
             : requestStub.responseBody.findIndex(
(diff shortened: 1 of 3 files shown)


packages/cypress-commands/src/setups/enableNetworkShim/captureRequests.js
```diff
@@ -166,8 +166,9 @@ function processDuplicatedRequest({
         // RequestStub was already nonDeterministic, responseBody is already an array
         const matchingResponseBodyIndex = isNotModified
             ? // When isNotModified the response body is empty and the browser is expected
-              // to return the last known response, so we do the same.
-              requestStub.responseBody.lenght - 1
+              // to return the last known response, so we do the same by returning the 
+              // value of the last responseLookup.
+              requestStub.responseLookup[requestStub.responseLookup.length - 1]
```
HendrikThePendric, line 171: This was wrong: For non-deterministic requests, the `responseBody` array is just a list of *unique* response bodies. The full sequence of requests, and which body should be returned when, is stored in the `responseLookup` array. So to get the last known response we should look at the value of the last item of the `responseLookup`.

packages/cypress-commands/src/setups/enableNetworkShim/index.js
```diff
@@ -44,25 +44,11 @@ export function enableNetworkShim() {
...
-                 * apart from the missing request stubs which do need to be
-                 * incrementally updated across features. So in stub mode we only
-                 * update that state property in the plugin.
-                 */
-                if (isStubMode()) {
-                    cy.task(
-                        'setNetworkShimMissingRequestStubs',
-                        networkShimState.missingRequestStubs
-                    )
-                }
```
HendrikThePendric, line 65: After fixing how fixtures were stored in `captureRequest`, stub-mode was still broken. It turned out this was because of the fact that the `responseCount` property of a `requestStub` was being cleared between tests. In order to preserve the value of `responseCount` between tests, we need to handle the state in exactly the same way as we do in capture mode so that it is incrementally updated between tests.

packages/cypress-plugins/src/plugins/networkShim/index.js
```diff
@@ -37,12 +37,6 @@ module.exports = function networkShim(
             state = newState
             return state
         },
-        setNetworkShimMissingRequestStubs(missingRequestStubs) {
-            if (Array.isArray(missingRequestStubs)) {
-                state.missingRequestStubs = missingRequestStubs
-            }
-            return state
-        },
```
HendrikThePendric, line 45: This has become redundant due to the changes in `packages/cypress-commands/src/setups/enableNetworkShim/index.js`

mediremi (approved):

HendrikThePendric:
> Before merging this in, I think I'd like to check what would happen if we apply these changes in isolation, without the changes from #261. I _think_ both PRs are required, but it could also be possible that the changes in the current PR are actually all that was needed.

I locally reverted the changes from #261 so that the fixtures were being uniquely identified by their full name (feature-name + scenario-name), but this broke the stub run of the reproduction case in the example platform-app. So we can conclude that both #261 and the current fix are needed.

PR #265, opened 2021-09-03
fix(install cmd): change setup call order & add network shim config

Mohammer5:
- [x] Calls `enableAutoLogin()` before `enableNetworkShim()`
- [x] Adds `"experimentalInteractiveRunEvents": true` to the `cypress.json` when adding the network shim


HendrikThePendric (approved):
LGTM.

One thing that didn't become immediately clear to me from simply looking at the changes is how this solved the problem of `enableNetworkShim()` being called before `enableAutoLogin()`. Intuitively I would have expected the options to be processed and inserted into the file in the same order as they are defined. As such, the changes in `options.js` seem wrong.

I wonder if perhaps it wouldn't make more sense to remove the call to `.reverse()` from here instead of reordering the options array:
https://github.com/dhis2/cli-utils-cypress/blob/6b83b9a8d6aa82294b260311cb5105d1ba20fd44/packages/cli/codemods/add-support-setup.js#L34

This will be functionally equivalent to what you've done here, so I am approving.

Mohammer5:
@HendrikThePendric I addressed what you've noticed, that was a good catch!

PR #297, opened 2021-11-09
fix(network shim): wait for network idle after tests

Mohammer5:
Background: Some of our tests were quite flaky. The cause is that some of the requests of the app are being fired when the test is actually done which causes them to not get recorded. When running cypress in stub mode, this produces an error as we're stubbing all requests of a certain url (which applies to requests that run during cypress' internal `after each` hook).

This PR adds a `waitForRequests` command that's being used before storing the captured requests.


HendrikThePendric:
Super happy you got to the bottom of this @Mohammer5! Well done 🎉

PR #378, opened 2022-12-07
chore: bump cypress to latest

HendrikThePendric:
I started this upgrade but it was a bit of a rabbit hole. The following points remain:
- [ ] In order to get the capture-runs in the platform app to succeed locally I was forced to add `@babel/helper-builder-react-jsx` and `@babel/helper-builder-react-jsx-experimental` to my package.json. There must be a better way to do this.
- [x] The auto-login feature is now broken. We were using `Cypress.Cookies.preserveOnce('JSESSIONID')` (in `packages/cypress-commands/src/setups/enableAutoLogin.js`) which is not compatible with the v12 API. I've updated this to `cacheAcrossSpecs: true,` in `packages/cypress-commands/src/commands/login.js`, but that starts causing issues on the second test. When visually inspecting the tests it actually seems like the login first succeeds and then a new attempt is done which fails.
- [ ] The most substantial point is the CLI package. We generate a lot of files. Pretty much all of them need to be changed according to the new naming scheme.

FYI @Mohammer5


Mohammer5:
Superseded by #393

PR #393, opened 2023-03-27
feat: remove cli util, update cypress to v12 & small conceptual adjustments

Mohammer5:
* No more `cy.login` (now there's a `cy.fillInLoginForm` command which can be used together with `cy.session`)
* The `d2-cli-cypress` tool is gone, it's not needed anymore (it is actually still there (the `install` command), but does nothing; as discussed with @HendrikThePendric)
* The commands & plugins libraries now have updated/correct peer dependencies
* Example apps now use cypress@12 & the latest versions of the commands / plugins
* Docs have been updated

## How to review this PR

There are several things that should be checked:

* Read the docs and ensure they're correct & understandable. I've changed quite a bit and I'd like the docs to be as good as possible. If there's something unclear, e.g. some context is missing, please let me know. In this case, too much is better than too little information.
* Ideally you use an app that has cypress test and follow the upgrade guide of the docs. That should ensure that the docs are complete & correct as well as that the libraries work as intended
* Repeat the steps I've listed below, under "How to test"
* Check if the `d2-utils-cypress install` command does nothing, just informing the user that the command is not supported anymore

### Example apps

* Check out this branch and install dependencies
* Run `yarn test:examples` (will start the example apps & run cypress tests, which includes both capturing and stubbing in the example app that uses the network shim)

### Real app

* Check out this branch and install dependencies
* Build the commands library (`yarn workspace @dhis2/cypress-commands build`)
* Link the commands & plugins libraries (`yarn workspace @dhis2/cypress-commands link && yarn workspace @dhis2/cypress-plugins link`)
* Check out the `upgrade-cypress-to-v12` branch of the aggregate data entry app (related PR: dhis2/aggregate-data-entry-app#319)
* Add the linked dependencies (`yarn link @dhis2/cypress-plugins @dhis2/cypress-commands`)
* Start the aggregate data entry app (`yarn cypress:start`)
* Play around with the UI (`yarn cypress:open:live`) and/or capturing/stubbing network requests (`yarn cypress:run:capture && yarn cypress:run:stub`)


docs/guides/using-the-network-shim.md
```diff
@@ -111,23 +95,25 @@ A workflow as described above consists of a few moving parts, which are describe
...
-    "video": false,
-    "env": {
-        "dhis2DataTestPrefix": "DATA_TEST_PREFIX",
-        "networkMode": "live", // default mode for the Network Shim
-        "dhis2ApiVersion": "value_from_prompt"
-    }
-}
+```js
+module.exports = defineConfig({
+    experimentalInteractiveRunEvents: true,
```
HendrikThePendric, line 104: I think we don't need this one anymore....

docs/guides/using-the-network-shim.md
```diff
@@ -149,42 +135,72 @@ This file is meant to be git-ignored and we can use this to set local environmen
...
+    // ...
 }
+
+// ...
 ```
 
-The `networkShim` plugin accepts a second `options` argument. For most DHIS2 platform apps the default options will be correct so there will be no need to pass a second argument.
+The `networkShim` plugin accepts a second `options` argument. For most DHIS2
+platform apps the options passed to `setupNodeEvents` will be correct so there
+will be no need to pass a second argument.
```
HendrikThePendric, line 162: I guess this should be *third* not *second* argument right? Also (pretty sure it was me who wrote this, so my apologies), but this text is not very clear at all. I probably meant to say that the _default options_ would be correct...

So this: 
> For most DHIS2 platform apps the options passed to `setupNodeEvents`

Should probably have been:
> For most DHIS2 platform apps the default options of the `networkShim`

Could you also verify that the above is correct? (i.e. that it's a third arg, and that the arg has some default options)

docs/guides/using-the-network-shim.md
```diff
@@ -149,42 +135,72 @@ This file is meant to be git-ignored and we can use this to set local environmen
...
 
-For most DHIS2 Cypress test suites the `enableAutoLogin` plugin can help reduce repetitive login steps in tests. In turn, the `enableAutoLogin` setup utility requires the `chromeAllowXSiteCookies` plugin when using Chrome. So a typical DHIS2 app will use this plugin alongside the `networkShim`.
+For most DHIS2 Cypress test suites the `enableAutoLogin` plugin can help reduce
+repetitive login steps in tests. In turn, the `enableAutoLogin` setup utility
+requires the `chromeAllowXSiteCookies` plugin when using Chrome. So a typical
```
HendrikThePendric, line 166: I think we should look into the value this plugin still has. I have some vague recollection of some Chrome updating breaking this. Have also found a [StackOverflow post](https://stackoverflow.com/questions/69150082/how-can-i-disable-same-site-cookies-flags-after-latest-update-of-chrome93) which seems to confirm this vague recollection.

packages/cypress-commands/src/setups/enableAutoLogin.js
```diff
@@ -1,22 +1,47 @@
...
+        /*
+         * At the very start of a capture run the server version
+         * is evaluated. This is done by querying the `system/info`
+         * endpoint which requires authentication. So if there is no
+         * valid session at the start of the run, we'd better log in.
+         */
+        cy.getCookie('JSESSIONID').then((cookie) => {
+            if (!cookie) {
+                createSession()
+            }
```
HendrikThePendric, line 40: I think this check could be problematic if we consider the session switching scenario. We probably need to work out another way to check if the session is valid already. I would actually hope that perhaps the Cypress API has our back...

HendrikThePendric (commented):
This is great stuff! There are a few points we need to discuss in the comments. And I also wonder how we should deal with removing a package. Probably we need to make sure we mention on NPM that it is deprecated.

docs/guides/using-the-network-shim.md
Mohammer5: > I guess this should be third not second argument right?

Don't think so:
image

> So this:
>
> > For most DHIS2 platform apps the options passed to setupNodeEvents
>
> Should probably have been:
>
> > For most DHIS2 platform apps the default options of the networkShim

Corrected

docs/guides/using-the-network-shim.md
Mohammer5: We talked about it and agreed to keep it as it doesn't hurt and won't ever change (no maintenance required) as Chrome/Chromium v94+ does not support this anymore.

I'm still using it as I have the latest Chrome but Chromium v93 installed on purpose as the developer tools are nicer to use than electron's dev tools.

PR #394, opened 2023-05-01
fix(auto login): ensure something has rendered in validation fn

Mohammer5:
Sometimes the `.should('not.exist')` check in the validate function was run before the app rendered anything, causing login issues.
By adding the change of this PR to the code inside `node_modules` in the aggregate data entry app's `upgrade_cypress_to_v12` branch, it then ran capturing & stubbing successfully consistently.


ismay (commented):
I've been using this to login:

```js
    cy.session(
        user,
        () => {
            // Login via API
            cy.request({
                url: `${user.server}/dhis-web-commons-security/login.action`,
                method: 'POST',
                form: true,
                followRedirect: true,
                body: {
                    j_username: user.name,
                    j_password: user.password,
                    '2fa_code': '',
                },
            })

            // Set base url for the app platform
            window.localStorage.setItem('DHIS2_BASE_URL', user.server)
        },
        {
            validate: () => {
                cy.request(`${user.server}/api/me`).then((response) => {
                    expect(response.status).to.eq(200)
                    expect(response.body.username).to.eq(user.name)
                })
            },
        }
    )
```

See: https://github.com/dhis2/scheduler-app/blob/fd8b877fbb735605495603557127f44397e3ef49/cypress/support/e2e.js

It's been a little faster for me than using the ui. Plus you're not dependent on rendering or any other FE parts. Would that be a solution you could use here?

Mohammer5:
I've implemented your suggestions but created commands for those particular tasks (logging in by api and validating a user session)

packages/cypress-commands/src/commands/validateUserIsLoggedIn.js
```diff
@@ -0,0 +1,6 @@
+Cypress.Commands.add('validateUserIsLoggedIn', ({ baseUrl, username }) => {
```
ismay, line 1: You could also call this validateSession maybe. Seems a little terser to me.

ismay (approved):

